PR #2, opened 2026-04-07
CLI-only interface: drop MCP as primary

henry-dowling:
## Summary
- Added 30+ missing CLI commands to reach full parity with the MCP server (91 CLI commands vs 82 MCP tools)
- New command groups: `files`, `docs`, `webhooks`, `sleep`, plus wiki features (`backlinks`, `graph`, `semantic-search`, `auto-index`) and table embeddings
- Updated CLAUDE.md to position CLI as the primary interface for both humans and agents (`boozle <cmd> --json`)

## Motivation
Three interfaces (MCP server, CLI, REST API) with no shared code is unnecessary complexity. The CLI can serve both humans and agents — agents call `boozle <cmd> --json` via Bash instead of MCP tools. One interface to maintain instead of three.

## Trade-offs
- Slightly slower for agents (process spawn per call vs persistent MCP connection) — negligible for human-paced interactions
- Less discoverability for agents (no auto-populated tool list) — mitigated by CLAUDE.md command reference
- MCP server is still in the repo but no longer the primary path

## Test plan
- [ ] `pip install -e .` and verify `boozle --help` shows all command groups
- [ ] Test new commands against live API: `boozle files list --json`, `boozle sleep config --json`, `boozle notebooks graph <id> --json`
- [ ] Verify agent workflow: Claude Code session using `boozle` CLI via Bash instead of MCP tools
- [ ] Confirm plugin hooks still work (they use their own lightweight client, unaffected)


samzliu:
@henry-dowling I removed MCP in my cutting fat exercise

PR #25, opened 2026-04-14
Strip to v0 OSS core — cut fat, zero LLM inference

samzliu:
**Summary**

  Major simplification pass that strips the codebase down to the essential v0 core: search, history, and wiki. Removes ~12k net lines of code.

  - Cut entire subsystems: MCP server, personas, chats/DMs, decks, documents, webhooks, realtime rooms, RAG client, sleep service (curation agent), and injection service
  - Zero LLM inference in backend: Moved all curation and search intelligence to plugin skills — the backend no longer calls any LLM APIs
  - Native Yjs converter: Replaced Node.js collab server dependency with a pure Python markdown ↔ Yjs XmlFragment converter
  - Flattened history: Merged history stores into a simpler flat model with new DB migration
  - UI overhaul: Merged workspace management into the home page, added dashboard visualizations (agent timeline, knowledge density map, embedding explorer), searchable member
  dropdowns, breadcrumb navigation on wiki pages
  - Misc fixes: WebSocket auth for Auth0 JWTs, robust key topics extraction, removed dead UI elements (URL image button, backwards-compat shims)

  **What was removed**

  - mcp_server/ — entire MCP server (~1,730 lines)
  - Routers: chats, decks, deck_viewer, dms, documents, personas, realtime, webhooks, search
  - Services: chat, deck, dm, document, persona_identity, ragflow_client, sleep, injection, webhook, watch, history_query, connection_manager, universal_search
  - Frontend pages: chats, decks, documents, personas, rooms, DMs, share dialog

PR #28, opened 2026-04-14
chore: drop dead /mcp/* Caddy handler

henry-dowling:
## Summary

PR #25 deleted the MCP server. No backend code serves \`/mcp/*\` anymore, so the reverse_proxy block in the Caddyfile pointed at a dead endpoint.

Verified:

\`\`\`bash
$ grep -rE 'mcp|MCP|/mcp' backend/
# no matches
\`\`\`

Stacks on #25. Retargetable to main once #25 lands.

## Scope note

This PR leaves \`/ws/*\` alone. That handler is removed in #26 (kill-yjs), where the WebSocket path itself goes away. Keeping these as separate PRs so each can land independently.

## Test plan

- [ ] \`caddy validate --config Caddyfile\` succeeds

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -16,10 +16,6 @@ app.example.com {
         reverse_proxy backend:3456
     }
 
-    handle /mcp/* {
-        reverse_proxy backend:3456
-    }
-
     handle /health {
         reverse_proxy backend:3456
     }


henry-dowling:
Superseded — #26 removed /mcp/\* along with /ws/\* in the Caddyfile cleanup, so this diff is already applied on cut_fat. Nothing left to do here.

PR #43, opened 2026-04-14
Port Octopus plugin to Cursor, Gemini, Codex, and opencode

henry-dowling:
## Summary

- Extracts agent-agnostic logic into `plugins/shared/` (HTTP client, hooks, event schema, state, tool-name normalization). Each per-agent plugin is a thin adapter + manifest + hook scripts.
- Adds four new plugins with real lifecycle hooks (not transcript scraping): **Cursor**, **Gemini CLI**, **Codex CLI**, **opencode**.
- Refactors the Claude plugin to use `plugins/shared/` — behavior preserved.
- Every event pushed to history is stamped with `metadata.client` (`claude_code` / `cursor` / `gemini_cli` / `codex_cli` / `opencode`) so one `agent_name` stays unified across editors while queries can slice by source tool.
- 25 parametric tests: adapter round-trip for every event kind across all 5 plugins + client-facet stamping.

**Prompt-time context injection is intentionally NOT part of this PR.** PR #48 gutted "agent identity + recent activity" injection on the Claude plugin; the multi-agent port keeps that deletion intact across all 5 plugins. Plugins are stream-only (plus auto-curation on SessionEnd for Claude).

Retrieval is CLI-based across all targets: each agent has shell-exec and can invoke `octopus history {query,search} --json` directly. See the Retrieval section in each plugin's README.

### Event coverage per agent

| Agent | session_start | prompt (stream) | tool_use | stop | session_end | Notes |
|---|---|---|---|---|---|---|
| Claude Code | ✅ | ✅ | ✅ | ✅ | ✅ | auto-curate via `claude -p` |
| Cursor | ✅ | ✅ | ✅ | ✅ (afterAgentResponse) | ✅ | — |
| Gemini CLI | ✅ | ✅ | ✅ (AfterTool) | ✅ (AfterAgent) | ✅ | — |
| Codex CLI | ✅ | ✅ | ⚠️ Bash-only | ✅ | — (uses Stop) | `notify` fallback for older builds |
| opencode | ✅ | ✅ | ✅ | ⚠️ (no stop signal) | ✅ (session.deleted) | TS shim spawns Python scripts |

## Test plan

- [x] `pytest plugins/tests/ --override-ini="addopts="` — 25/25 pass
- [x] Claude plugin smoke-tested end-to-end after refactor (session_start + tool_use scripts)
- [x] Each per-agent plugin smoke-tested on sample stdin
- [ ] Real end-to-end test inside Cursor, Gemini CLI, Codex CLI, opencode — **needs the reviewer's help, no account/session handy in dev**

## Caveats

- Stdin payload shapes for Cursor/Gemini/Codex are my best reading of public docs as of 2026-04. Adapters tolerate missing fields gracefully and `plugins/tests/fixtures/` is the single source of truth when upstream shape changes.
- Codex tool hooks are Bash-only today; `on_notify.py` covers non-Bash turns via the stable `notify` array. Both are wired up and dedup server-side.
- opencode has no clean "turn complete" signal — `session.idle` fires on every turn; final-assistant-message capture per turn is a future TODO.


henry-dowling:
**Terminology: `plugins/shared/` should say "client," not "agent"**

This PR introduces the split itself:

- `agent_name` (existing — `backend/models.py`, `memory_service.py`, CLI `--agent`) = the registered agent identity, unified across editors.
- `metadata.client` (new in this PR) = the source tool (`claude_code` / `cursor` / `gemini_cli` / `codex_cli` / `opencode`).

The runtime code in `shared/hooks.py` + `shared/octopus_client.py` respects that — `cfg["agent_name"]` and `cfg["client"]` are passed as separate fields.

But the docstrings/comments in `plugins/shared/` (and the PR body) use "agent" to mean the host editor, which overloads `agent_name`:

- `plugins/shared/__init__.py`: "across all coding agents" / "per-agent plugin"
- `plugins/shared/event.py`: "shared across all agent plugins" / "its agent's raw stdin JSON"
- `plugins/shared/hooks.py`: "Agent-agnostic hook logic" / "any specific agent's payload shape"
- `plugins/shared/state.py`: "Each agent plugin passes in its own data_dir"
- `plugins/shared/summarize.py`: "which agent called it"

Suggest renaming the vocabulary in these docstrings (and the per-plugin READMEs where it leaks in) to **client** / **per-client plugin** / **client-agnostic** so the prose matches the schema this PR just established. Otherwise a reader who's seen the DB schema will parse "per-agent plugin" as "one plugin per `agent_name`" — which is the opposite of what's happening.

henry-dowling:
Findings from PR43 review:

1) P1 (blocker): `stream_stop` is emitting `session_id` from state instead of the event payload on stop-streaming paths.
- In `plugins/shared/hooks.py`, both `assistant_message` and `session_end` events are constructed with `session_id=state.get("session_id", "")`.
- `on_notify` and `on_stop` flows can call this with an event that already has `event.session_id` (including non-empty, threaded notify sessions), but that ID is ignored.
- This can mis-associate summaries and end-events, especially if state is stale or cleared.

2) P3 (cleanup): There is lingering terminology drift (`agent` vs client/metadata) in shared docs/comments (`plugins/shared/__init__.py`, `plugins/shared/event.py`, `plugins/shared/state.py`, `plugins/shared/summarize.py`). Not functional, but worth aligning.

Recommendation: pass through `event.session_id` for stop-streaming events (with clear fallback behavior only when missing).

plugins/claude-plugin/skills/connect/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: connect
-description: Set up your Octopus agent name and workspace. Use when starting with Octopus for the first time or reconfiguring the connection.
+description: Set up your Octopus agent identity, workspace, and history store. Use when starting with Octopus for the first time or reconfiguring the connection.
```
henry-dowling: The frontmatter description here adds "history store" but the rest of this same commit removes that concept (step 6 now reads *"Events stream directly to the workspace memory — no separate history store"*, and `history_store_id` checks were dropped from `shared/hooks.py`). Suggest reverting this line to something like *"Set up your Octopus agent identity and workspace."*

henry-dowling:
## Follow-up work

Items from the adversarial review that were deliberately deferred out of this PR. Filing here so they don't get lost. Rough priority by risk × effort:

### Medium — should land before external users

**1. Lost-update race on `state.json`** — `_write_json` is atomic (pid-suffixed tmp + `os.replace`), but the `load_state → mutate → save_state` pattern in `on_stop.py` / `on_session_end.py` / `on_session_start.py` is not. If two hook scripts in the same session interleave (opencode fires many in parallel; Claude SessionEnd racing Stop is also possible), one writer's `session_id` update can be clobbered by another writer that `load`ed before the first `save`. Fix: `fcntl.flock` on a sidecar `~/<data_dir>/state.lock` around each load-modify-save pair, or model state as separate files per concern (`session_id.json`, `streaming.json`). ~8 call sites to update.

**2. PII / secret redaction in `summarize.py`** — `bash` tool summaries ship 300 chars of command + 500 chars of stdout + 200 chars of stderr to the backend. `Edit` ships 100 chars each of `old_string` / `new_string`. That's the feature, but there's no redaction layer — if the user's agent edits `.env`, runs `aws secrets get-value`, or diffs a file containing an API key, it all ships. Base_url is locked to `~/.octopus/config.json` now, so the data only goes to the user's legitimate backend — but operators of that backend then store it in plaintext. Options: (a) regex redaction at the plugin boundary (AWS/GH/Slack token shapes, common env-var patterns), (b) opt-in allowlist instead of opt-out, (c) hash-and-store-locally for tools that are high-risk, (d) document the threat model clearly and punt to backend-side redaction.

**3. Lost-update on central `~/.octopus/config.json`** — same race as #1 but for `record_curate_run`, `set_streaming_enabled`, `set_auto_curate`. `_write_central` does `_read_central + update + _write_json` with no lock. Multiple plugins writing concurrently (e.g. two Claude sessions ending at once, or opencode's parallel hook fan-out) can lose a `last_curate_at` update. Same flock fix applies.

### Low — nice to have

**4. `auto_curate` truthiness drift in per-plugin config** — the central `auto_curate_enabled()` handles `"1" / "yes" / "true" / bool`, but each `config.py` still defaults to the string `"true"` / `"false"` and compares via `str(...).lower() != "true"` in a couple places. Consolidate on the central helper everywhere. Cosmetic.

**5. Truncation marker in metadata** — prompts clip at 2000 chars (`hooks.py:28`), assistant messages at 4000 (`hooks.py:91`), bash outputs at 500 (`summarize.py:41`). Downstream consumers can't tell "short message" from "truncated prefix." Add `metadata["truncated"] = True` when we clip. 5 lines.

**6. `client` facet validation at push-time** — today every `config.py` hardcodes a literal string (`"cursor"`, `"codex_cli"`, etc.) into `cfg["client"]`, so it's not attacker-controlled. But `OctopusClient.push_event` takes it as an opaque string. If a future contributor wires `client` through `_cli_config()` or env vars, it becomes attacker-controlled metadata. Lock to a hardcoded enum at push time. 3 lines + one test.

**7. opencode final-assistant-message capture** — documented in `opencode-plugin/README.md` as a known gap. `session.idle` fires per-turn and has no text; `message.updated` streams partials. A clean fix probably requires listening for a specific part-type boundary in the opencode event bus. Low priority until someone asks for it.

**8. Gemini/Codex transcript stats** — `stream_session_end` only emits tool_count / files_changed for Claude because the transcript format is Claude-specific. For Gemini and Codex the session_end summary always says "Session ended." with no stats. Either wire per-adapter transcript parsers (one per agent) or accumulate stats in state during the session.

---

None of the above is blocking this PR IMO. Filing as a checklist to triage. Happy to open individual issues for any of these if that helps tracking.

PR #49, opened 2026-04-15
fix: login redirect to workspace + clean up page graph

samzliu:
## Summary
- **Login redirect**: After login, fetches user's workspaces and redirects to the single workspace homepage if they have exactly one, otherwise to the workspace manager (`/`). Previously always redirected to `/memory`.
- **Page graph cleanup**: Removed grey text labels drawn over graph nodes (hover tooltip still shows names) and removed the close button from the inline graph view.
- **Dashboard viz overhaul**: Overhauled dashboard visualizations (carried from prior work).

## Test plan
- [ ] Log in with a user that has 1 workspace — should land on `/workspaces/{id}`
- [ ] Log in with a user that has 0 or 2+ workspaces — should land on `/`
- [ ] Open page graph in a notebook — nodes should not have text labels, only hover tooltips
- [ ] Inline page graph should not show a close button

### DIFF
--- a/backend/migrations/versions/0006_drop_auth0_sub.py
+++ b/backend/migrations/versions/0006_drop_auth0_sub.py
@@ -3,12 +3,12 @@
 Auth0 integration is removed; login is password + API key only.
 
 Revision ID: 0006
-Revises: 0005
+Revises: 0001
 """
 from alembic import op
 
 revision = "0006"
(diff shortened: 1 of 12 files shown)


henry-dowling (approved):
Great call on the redirect, lgtm + I'm gonna merge it

PR #75, opened 2026-04-16
fix(frontend): relocate managed overlay into frontend/

henry-dowling:
## Summary
- stash-web Render deploys have been failing since PR #73 landed: `Module not found: Can't resolve '@managed/auth0/*'` during `npm run build`.
- Render builds the frontend with `dockerContext=./frontend`, so the `managed/` overlay at the repo root was never copied into the image. Turbopack additionally refuses to resolve `@managed/*` imports that walk above the Next.js project root, so copying the overlay in beside `/app` would not have worked either.
- Move `managed/frontend/auth0/*` → `frontend/managed/auth0/*` and point the alias at `./managed/*`. No other package referenced the old path, so the move is contained.

## Test plan
- [x] `docker build -f frontend/Dockerfile frontend` succeeds locally
- [ ] Render deploy for stash-web goes green on merge


henry-dowling:
Superseded by #76 (same fix, rebased so it merges cleanly against the current main).

PR #115, opened 2026-04-17
fix(www): step 5 — render welcome verbatim instead of punting to /stash:welcome

henry-dowling:
## Summary

Step 5 of the install prompt previously said *"Tell me to run \`/stash:welcome\` to see the onboarding."* — that reflected the architectural reality (Claude can't invoke slash commands; only the user can type them) but pushed the action onto the user, who often missed the cue.

This reverts step 5 to \`stash connect --welcome\` with an explicit instruction to **paste the output back as a markdown block in the assistant's reply**, so the markdown renders inline instead of being buried in Claude Code's collapsed Bash-result UI. The \`/stash:welcome\` slash command (#112) stays useful for re-reading later.


henry-dowling:
🎉 Confirmed working end-to-end as a new user — this is the first commit where the entire landing-page install prompt runs cleanly without manual intervention. Install + plugin + sign-in + workspace auto-default + welcome-rendered-inline all firing as designed.

PR #121, opened 2026-04-17
fix(install): use bash -c form, drop the /dev/tty hack (py3.14 compat)

henry-dowling:
## Summary

Reported by @htdowling: the \`</dev/tty\` redirect from #120 got past the "not a terminal" error but Python 3.14's asyncio kqueue can't \`add_reader\` on the redirected fd, so questionary crashes with \`OSError: [Errno 22] Invalid argument\`.

\`\`\`
File "selectors.py", line 505, in register
  self._selector.control([kev], 0, 0)
OSError: [Errno 22] Invalid argument
\`\`\`

Verified locally that kqueue + add_reader works fine on a regular pipe fd under py3.14 — the bug is specific to the redirected /dev/tty fd.

The reliable fix is the \`bash -c "\$(curl …)"\` invocation (used by rustup, brew, etc.). In that form the script body is argv, not stdin, so stdin stays attached to the user's real terminal and there's nothing to redirect.

- \`install.sh\` drops the /dev/tty redirect; when stdin still isn't a TTY (someone uses the legacy \`curl | bash\` form or runs in CI), it prints the recommended invocation + exits 0 instead of crashing inside questionary.
- Landing-page hero \`INSTALL_COMMAND\` swaps to the \`bash -c\` form.

New canonical install:

\`\`\`bash
bash -c "\$(curl -fsSL https://raw.githubusercontent.com/Fergana-Labs/stash/main/install.sh)"
\`\`\`

## Test plan

- [x] \`bash -n install.sh\` syntax-checks; \`npm run build\` clean
- [x] Repro: kqueue add_reader fine on pipe under py3.14
- [ ] Reviewer: re-run with the \`bash -c\` form on a fresh box

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,8 +3,15 @@
 # `stash connect` questionnaire (scope, managed-vs-self-host, browser auth,
 # workspace, agent plugin).
 #
-# Usage:
-#   curl -fsSL https://stash.ac/install | bash
+# Recommended invocation (keeps stdin attached to your terminal so the
+# questionnaire's interactive picker works):
+#
+#   bash -c "$(curl -fsSL https://raw.githubusercontent.com/Fergana-Labs/stash/main/install.sh)"
+#
+# `curl … | bash` works for the install part but the interactive picker
+# needs a real terminal on stdin — bash's `| bash` makes the script body
+# stdin instead. We try to detect that case and surface an actionable
+# message rather than crashing inside questionary.
 #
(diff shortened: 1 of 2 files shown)


henry-dowling:
🎉 **Onboarding works.** First commit where the full one-liner installs cleanly end-to-end on a fresh box: bash -c "$(curl …)" → questionnaire → browser sign-in → workspace auto-default → Claude Code plugin install → done.

PR #128, opened 2026-04-17
feat: evaluation system + retrieval & scoring fixes driven by eval results

triobaba:
## Summary

- **Built a comprehensive evaluation system** (`evals/`) with 6 suites covering the full memory pipeline
- **Used the eval system to identify and fix two critical production defects** in scoring and retrieval
- **Eval results improved dramatically** after fixes: NDCG@5 from 0.041 to 0.774

### Architecture

The eval system is a self-contained Python package (`evals/`) that spins up isolated test data in the real Postgres database, runs Octopus's actual `injection_service` and `notebook_service` against it, and measures results with information retrieval metrics. No mocks on the retrieval path — this tests the real pipeline.

```
evals/
├── datasets/          # JSONL scenario files (hand-authored)
├── judges/            # Metrics (NDCG, MRR, P@k, R@k) + LLM-as-judge
├── suites/            # 6 evaluation suites
├── reports/           # Console (Rich) + JSON baseline comparison
├── harness.py         # Core test harness (DB setup, page insertion, injection runner)
├── config.py          # Thresholds and weights
├── conftest.py        # Pytest integration
└── __main__.py        # CLI: python -m evals --suite <name>
```

### Suite 1: Retrieval (15 scenarios)

Tests FTS retrieval accuracy in isolation. Each scenario defines a query, a target page that should be retrieved, and distractor pages that shouldn't outrank it. Measures **NDCG@5, MRR, Precision@k, Recall@k**. Scenarios cover API docs, architecture decisions, deployment configs, debugging notes — realistic notebook content.

**Approach**: Insert target + distractors into a fresh notebook, run `search_pages_fts`, check if the target appears at rank 1 and compute IR metrics against the ideal ranking.

### Suite 2: KG Relations (8 scenarios)

Tests the knowledge graph relation extraction pipeline. Each scenario provides page content and expected relations (e.g., `depends_on`, `contradicts`, `supersedes`). Verifies that `upsert_relation` correctly stores typed edges and that `invalidate_conflicting_relations` handles contradictions.

**Approach**: Insert pages, call relation functions, query `page_relations` table, check relation types and directionality.

### Suite 3: Sleep Agent (5 scenarios)

Tests the sleep agent's curation executor — the function that takes LLM-generated actions (`create_page`, `update_page`, `extract_relations`) and applies them to the notebook. Uses mocked LLM responses to isolate executor logic from LLM variability.

**Approach**: Feed pre-defined action sequences to the executor, verify pages are created/updated correctly and relations are stored. Optional LLM-as-judge scoring for output quality when `ANTHROPIC_API_KEY` is set.

### Suite 4: End-to-End (8 scenarios)

Full pipeline test: insert pages → run `compute_injection` (the actual injection service) → check if the right pages are selected for context injection. Tests the complete flow including FTS search, scoring, and page selection.

**Approach**: Each scenario defines a user prompt, notebook pages, and expected pages that should appear in the injection. Runs the real `compute_injection` function and checks hit rate.

### Suite 5: Degradation (9 scenarios, 3 types)

Stress tests that expose how retrieval degrades under real-world pressure:

- **Scale degradation** (3 scenarios): Inserts a target page alongside 10, 100, 500, and 1000 distractor pages, then queries for the target. Measures NDCG@5 at each scale level and fails if the drop from smallest to largest exceeds 30%. This catches bloat-induced retrieval collapse.

- **Temporal suppression** (3 scenarios): Inserts an "always-inject" page and runs 25 consecutive injection rounds with `run_injection_loop` (which chains session state between rounds, simulating a real multi-turn session). Checks that the critical page survives recency decay and reappears across rounds — catches the failure mode where important pages get buried by the recency score.

- **Near-duplicate disambiguation** (3 scenarios): Inserts multiple versions of the same page (e.g., v1 through v5 of "API rate limits") with slightly different content and timestamps. Queries for the topic and checks that the most current version ranks highest — catches the failure mode where stale duplicates outrank the latest truth.

### Suite 6: Curation Quality (5 scenarios)

End-to-end test of the full curation-to-retrieval loop: history events → sleep agent curation → notebook pages → retrieval query. This is the most comprehensive suite because it tests the entire memory pipeline.

**Approach**: Each scenario provides raw history events (e.g., "Implemented Redis caching for user session tokens") and retrieval queries with expected keywords. The harness feeds history to the sleep agent (with mocked LLM responses), lets it create notebook pages, then runs retrieval queries against the created pages and checks if expected keywords appear in the results.

### Metrics & Tooling

- **IR Metrics**: NDCG@5, MRR, Precision@k, Recall@k, F1 — implemented from scratch in `judges/metrics.py`
- **LLM-as-Judge**: Claude-based qualitative scoring with structured rubrics (relevance, completeness, actionability) in `judges/llm_judge.py`
- **Baseline Comparison**: JSON reports with delta tracking — `python -m evals --compare baseline.json` shows regressions
- **CI Integration**: All suites in `test.yml` on push/PR; nightly workflow for degradation + curation quality

---

### Fix 1: Scoring formula (multiplicative → additive weighted)

The old `score = relevance × recency × staleness × confidence` silently zeroed out highly relevant pages. Example: a page with relevance=0.9 but recently shown (recency=0.1) scored 0.0135, while a mediocre page (relevance=0.3) scored 0.24.

New: `score = 0.50·relevance + 0.15·recency + 0.15·staleness + 0.20·confidence`

The relevant page now correctly outranks (0.69 vs 0.62). **Identified by**: retrieval eval suite returning near-zero NDCG scores despite correct FTS matches.

### Fix 2: FTS retrieval (AND → OR + trigram fallback)

`search_pages_fts` used `websearch_to_tsquery` requiring ALL query terms to match. A query like "API rate limits for each pricing tier" produced `'api' & 'rate' & 'limit' & 'price' & 'tier'` — but if "price" wasn't in the page, the entire page was invisible despite matching 4/5 terms.

Fix: OR-connected tsquery via `plainto_tsquery` with `&` → `|` replacement, plus `pg_trgm` `word_similarity` fallback for stemming mismatches. New migration `0004_add_pg_trgm.py`.

**Identified by**: 13/15 retrieval scenarios returning 0 results in the initial eval run.

---

## Eval Results (before → after fixes)

| Suite            | Before     | After        | Delta  |
|------------------|------------|--------------|--------|
| retrieval        | 2/15 (13%) | 13/15 (87%)  | +73%   |
| end_to_end       | 2/8  (25%) | 8/8  (100%)  | +75%   |
| degradation      | 5/21 (24%) | 19/21 (90%)  | +67%   |
| curation_quality | 0/5  (0%)  | 4/5  (80%)   | +80%   |
| NDCG@5           | 0.041      | 0.774        | +0.733 |
| MRR              | 0.067      | 0.800        | +0.733 |

Scale test: NDCG@5 = 1.0 at 10, 100, 500, and 1000 pages (0% degradation).
Temporal test: always-inject pages survive 100% of 25 injection rounds.

Both pre-fix and post-fix JSON baselines are included in `evals/reports/output/`.

### Remaining 3 failures (known architectural limits)

- 2 retrieval scenarios need vector/embedding search (semantic similarity beyond FTS)
- 1 curation quality scenario needs `created_at` boost for temporal ordering

These are out of scope for this PR and tracked as future work.

---

## CI

- All 6 suites run on push/PR in `test.yml`
- Nightly workflow (`eval-nightly.yml`) runs degradation + curation quality with baseline comparison

## Test plan

- [ ] `python -m evals --suite retrieval` passes ≥ 80% scenarios
- [ ] `python -m evals --suite degradation` shows < 30% NDCG drop at 1000 pages
- [ ] `python -m evals --suite curation_quality` passes ≥ 60% hit rate
- [ ] CI workflows run successfully on push
- [ ] Existing backend tests unaffected


henry-dowling:
Thanks so much for the thorough POR! Unfortunately I'm gonna close this because it's gotten too far behind main to review or merge (you can see that the diff is ~65k lines!) We'd like you to make a new PR off of main.
 
Furthermore, we're concerned that this PR is answering the wrong question for where we are in our development right now- it answers "does our implementation of our approach to memory work?" rather than "is our approach to memory the right approach?" 

Some notes from @samzliu 

The goal here is to build a lightweight "benchmark" as opposed to a "test" or "eval". See https://hai.stanford.edu/policy/what-makes-a-good-ai-benchmark - while we don't want a full benchmark, we do want to build a way to evaluate any AI memory system we build on the dimensions we care about.
    - Test - (usually deterministic) checks to make sure your code is working
    - Eval - Answers: Is our system performing up to the standard that we want it to?
    - Benchmark - Where are we in terms of STOA capabilities on this metic we care about?
- As mentioned, we also want to make sure we are able to cheaply and quickly test this
- On Friday, we discussed the two main problems we want to measure and improve against are:
    - long term stability
    - relevance / blast radius for a retrieved memory
- For long term stability, I imagine something like the following would probably work well: 
    - We have a way to measure how much "slop" there is in a memory store
    - We then have a "checkpoint" of a memory store which has low slop
    - We then introduce a number of new things to insert (both distractors, relevant things, etc.) and run our janitor/sleep time agent
    - And measure slop after
- For blast radius,
    - I think a similar mechanism to your retrieval system could work well. The key here is probably to develop curated examples where retrieval fails rather than raw retrieval against distractors (which is not the exact failure mode that we see in real life). It should be more - can the model in XYZ situation tell what are right memories to apply (similar to the memory as prediction article)

PR #174, opened 2026-04-20
Manifest-only streaming scope; ask before opening codex sandbox

henry-dowling:
## Summary

Two changes to how hooks decide whether and where to stream — they landed together because they're the same mental-model cleanup on two sides (config + onboarding).

### 1. Streaming gate: manifest presence is the only signal

Before, three things were competing to express "is this session in scope?": the `scope` central-config field (`repo|workspace|all`), the `install_repo_common_dir` single-slot pin, and `.stash/stash.json` manifests. They could disagree, the UI promised workspace-level gating that was never implemented, and multi-repo installs required contorting the single-slot pin.

After:
- `stashai/plugin/scope.py` — `cwd_in_scope` is a walk-up for `.stash/stash.json`. Nothing else.
- Per-repo install: commit `.stash/stash.json` at the repo root (already what `stash connect` does).
- Global install: `stash install --global` writes `~/.stash/stash.json`. Since the walk-up climbs from cwd through `$HOME`, every session under home streams — unless a closer repo manifest overrides routing.
- Closest-ancestor wins for routing. Repo manifest beats global manifest automatically.

Deleted: `SCOPE_CHOICES`, `_capture_install_repo`, scope row in `stash settings`, `stash config scope` handler, `get_scope_config`, `install_repo_common_dir` reads/writes.

### 2. Codex install prompts before opening the sandbox

Codex's default `workspace-write` sandbox blocks outbound network from bash commands, silently killing stash hook uploads. Earlier the install silently added `[sandbox_workspace_write] network_access = true` to `~/.codex/config.toml` — a posture change hidden inside an idempotent config append.

Now `stash install codex` asks before writing the block (via `_ask_codex_network_access`). If the user declines, the top-level block is stripped from the snippet but `[profiles.stash]` stays, so `codex --profile stash` continues to work as an escape hatch. Non-TTY installs default to yes so scripted installs keep working.

## Test plan

- [x] `python -m pytest plugins/tests/ cli/tests/ --no-cov` — 68 passing.
- [x] Asset-in-sync test confirms the codex snippet stays byte-identical between `plugins/codex-plugin/` and `stashai/plugin/assets/codex/`.
- [x] Manifest walk-up tested: in-cwd, in-ancestor, missing, nested-precedence.
- [x] Install upgrade paths tested: fresh, missing-both, missing-only-sandbox, idempotent re-run.
- [ ] Manual: run `stash install codex` interactively in a fresh env, confirm prompt fires. *(needs a clean machine — safe to skip if CI green.)*
- [ ] Manual: run `stash install --global` and verify `~/.stash/stash.json` is written with correct workspace_id.

## Migration notes

- **`scope=repo` users** (most): behavior is unchanged if their repo has a `.stash/stash.json` manifest. If not (edge case: older connect runs that pre-date `_offer_repo_init`), re-run `stash connect` in the repo to drop one.
- **`scope=all` users**: run `stash install --global` to write a home-level manifest, then the old all-bypass behavior returns. Until they do, streaming will no-op outside of manifest trees.
- **`scope=workspace` users**: this was a silent alias for `all`. Same migration as `all`.
- **`install_repo_common_dir`**: no longer honored. Run `stash connect` in the repo to write its manifest.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -409,6 +409,47 @@ def _upsert_agents_md(path: Path, body: str) -> None:
     path.write_text(new)
 
 
+def _ask_codex_network_access() -> bool:
+    """Prompt the user to enable top-level `network_access` for codex's
+    workspace-write sandbox. Defaults to True on non-TTY so batch installs
+    don't hang."""
+    if not sys.stdin.isatty():
+        return True
+    console.print(
+        "\n[bold]Codex sandbox network access[/bold]\n"
+        "Codex's default sandbox blocks outbound network from bash commands.\n"
+        "Stash hooks use bash to POST session events to api.stash.ac, so they\n"
+        "need network to upload. Enabling this lets plain `codex` stream.\n"
+        "You can opt out now and use `codex --profile stash` when streaming,\n"
+        "or edit ~/.codex/config.toml later to change your mind."
+    )
+    try:
+        answer = questionary.confirm(
+            "Allow codex bash commands to make outbound network requests?",
+            default=True,
+        ).ask()
+    except Exception:
+        return True
+    return True if answer is None else bool(answer)
+
+
+def _strip_top_level_sandbox(snippet: str) -> str:
+    """Remove the `[sandbox_workspace_write]` top-level block (and its
+    explanatory comment) from the snippet, leaving `[profiles.stash]`
+    and everything else intact."""
+    start = snippet.find("[sandbox_workspace_write]")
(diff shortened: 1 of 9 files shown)


henry-dowling:
Split into two atomic PRs for easier review:

- #175 — codex onboarding (ask before opening sandbox for network)
- #176 — scope refactor (manifest is the only streaming opt-in)

PR #237, opened 2026-05-07
fix: public workspaces copy share link instead of invite link

henry-dowling:
## Summary
- Public workspaces now copy `/s/<workspaceId>` (no auth required) instead of `/join/<code>` (requires login)
- Button label changes to "Copy Share Link" for public workspaces
- Private workspaces still copy the invite link as before

## Test plan
- [ ] Open a public workspace, click the button — link should be `/s/<id>` format
- [ ] Open a private workspace, click the button — link should still be `/join/<code>`
- [ ] Visit the copied public link in an incognito window — should load without login

### DIFF
--- a/frontend/src/components/workspace/WorkspaceSidebar.tsx
+++ b/frontend/src/components/workspace/WorkspaceSidebar.tsx
@@ -160,8 +160,10 @@ export default function WorkspaceSidebar({
     }
   };
 
-  const copyInvite = async () => {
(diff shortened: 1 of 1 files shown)


henry-dowling:
Closing this as no longer needed. The workspace ShareSheet from PR #235 already mints and copies the public /s/<workspaceId> link, while the existing invite button can remain focused on /join/<invite_code>.

PR #252, opened 2026-05-12
Optimize session detail loading and sidebar spine data

henry-dowling:
## Summary

Session-navigation speed was regressed bc we were requesting way more session titles than we needed for the navbar. We added the ability to limit the number of session titles returned by a request. This change set now adds request-scoped spine caching so shared UI surfaces can reuse spine payloads across transitions. we cache in memory, not local storage btw

## What changed

- Backend: added parameters to `GET /api/v1/stashes/{stash_id}/spine`
  - `session_limit`
  - `wiki_depth` (`full|root`)
  - `include_file_urls`
  - response now includes `session_count`
- Backend: added focused session-spine query optimization in `memory_service.list_workspace_sessions` and new migration `0036_session_spine_indexes.py` for history events performance.
- Backend: transcript endpoint now returns transcript metadata + events in one request, avoiding a second query from session detail page.
- Frontend:
  - Sidebar/Command Palette now request constrained spine payloads for navigation search.
  - Session detail page now uses single `/events` call for both metadata and event stream.
  - Stash home uses constrained spine request with root/full wiki behavior adjusted where needed.
- Frontend cache improvement:
  - Added in-memory shared cache + in-flight dedupe for `getStashSpine` in `frontend/src/lib/api.ts`.
  - Repeated calls with identical `stashId` + spine query options now reuse the same response within a short TTL (30s), reducing duplicate network requests while navigating between session/home/palette surfaces.

## Why

Session navigation and session detail now avoid loading unnecessary data and avoid repeated duplicate spine calls for equivalent payload scopes, reducing latency in large stashes.

## Validation

- Not run in this pass (can be run if you want).

## Notes

- Migration introduced: `backend/migrations/versions/0036_session_spine_indexes.py`.

## Screenshots

- This update is performance-focused with minimal UI behavior changes; screenshots can be attached in follow-up review if needed.


henry-dowling:
Updated screenshots for the unified Wiki navigation changes in `100c4d5`:

Stash home showing Sessions and one Wiki section

Command palette using page/folder/session/file language

PR #278, opened 2026-05-12
Stash handoff curator + server-side LLM unification

samzliu:
## Summary

- **New: per-stash handoff document.** Agent-written markdown that orients a coding agent landing on a stash — what's here, what's going on, where to start, and a human-written Stash Description. Lives on the stash home page and behind a new `stash handoff show` / `stash handoff refresh` CLI.
- **Fresh-or-nothing.** `GET /handoff` returns the body only when it's fresh (or pinned by a human). Stale or never-generated rows report `{present: false, reason}` so callers never act on out-of-date orientation. `POST /handoff/regenerate` is synchronous (advisory-lock + 300s).
- **Sleep-time writer worker.** Daily-cadence regenerator with 5-min write debounce + fingerprint short-circuit. Runs the same toolset as ask-the-stash so the writer can explore, not just summarize a seed. Hard caps prevent unbounded cost.
- **Stash Description**, not principles wiki page. The `workspaces.description` column is now the canonical human-written input for the writer agent. Edit it inline on the stash home page; the `HANDOFF_PRINCIPLES.md` lookup is gone.
- **Server-side session summarizer.** Plugin uploads artifacts only. New sessions default to `summary_status='need_summary'` and the backend worker claims them atomically (`UPDATE … SET summary_status='in_progress' WHERE summary_status='need_summary'`).
- **`sessions.summary_status`** with a real enum: `need_summary | in_progress | failed | done` (was `live | summarizing | ready | failed`, which conflated session lifecycle with summarization state).
- **`/spine` → `/overview` + new `/sidebar`.** The stash home loads `{handoff_metadata, sessions, wiki}` from `/overview`. The nav tree loads only what it needs from `/sidebar`, with `ETag` + `Cache-Control` so navigation hits 304 instead of refetching.
- **Claude Agent SDK** replaces the hand-rolled agent harness. Ask-the-stash, the handoff writer, and the session summarizer all go through `backend/services/agent_runtime.py`. The 8 in-process tools become a single in-process MCP server, workspace-scoped via `ContextVar` so tools can find the active stash without threading the id through every call.
- **CLI:** `stash handoff show` and `stash handoff refresh` (no more `stash stash handoff`, no more `--wait` / `--refresh` flags).
- **Pin UI, restrained.** A small `📌 Auto-curation off` badge next to the panel header + a `Turn auto-curation back on` action. No full-width banner.

## Why

A new agent landing on an unfamiliar stash shouldn't have to grep. The handoff makes that startup cost zero — but only if we can trust it. The earlier "force-regen with `?wait=true`" path put freshness on the caller; this version makes that the writer's job, and `GET /handoff` either returns trustworthy content or admits it doesn't have any.

The SDK swap is the bigger structural change: the old hand-rolled agent loop in `backend/services/llm.py` was ~210 lines of tool-dispatch + token caps that the official SDK now owns. Less code, same toolset, same per-stash scoping.

The summarizer/status split fixes a subtle ownership bug. Before, the plugin set `status='summarizing'` and the worker filtered on `('live','summarizing')` — overlapping concerns. Now the worker claims atomically with `UPDATE … RETURNING`; if a worker dies, the row stays in `need_summary` and the next worker picks it up.

## Verification

- [x] Migration 0038 + 0039 round-trip clean. `summary_status` carries the new enum; `idx_sessions_needs_summary` is rebuilt against the new value.
- [x] Backend imports clean; `ast.parse` + `import` smoke-test passes for every touched module.
- [x] Targeted ruff + black on every touched Python file.
- [x] Frontend `tsc --noEmit` passes.
- [ ] **End-to-end smoke (post-merge, staging, with `ANTHROPIC_API_KEY`):**
  - Create a stash, edit the Description inline.
  - Run an agent → confirm `summary_status` transitions `need_summary → in_progress → done` within ~10s of the session ending.
  - Confirm the writer regenerates the handoff within 5min, surfacing the Stash Description in the rendered body.
  - `GET /handoff` returns `{present: false, reason: 'stale'}` immediately after editing a wiki page; once the writer runs it flips to `{present: true, reason: 'fresh'}`.
  - Edit the handoff in the UI → `📌 Auto-curation off` shows; click _Turn auto-curation back on_ → next tick regenerates.
  - `stash handoff show` / `stash handoff refresh` print the expected output. No `stash stash handoff` form exists.

## Screenshots

Will attach screenshots of the HandoffPanel (empty / present / pinned) and the inline Stash Description editor in a follow-up comment once a dev DB has migration 0039 applied and a real `ANTHROPIC_API_KEY` is configured.


henry-dowling:
Okay, here are a lot of thoughts inspired by reading this PR. Some of these are thoughts that don't directly apply to the PR but rather just occurred to me as I began to understand our system more deeply while reviewing this PR:

## Directly relevant to this PR:

1. Why have a forced handoff regeneration path? It should not be on the caller to know the internal freshness state of the handoff. We should just promise them that they'll get a good handoff document every time they ask for one. I don't really see the utility of ever giving them an out-of-date handoff document- that feels worse than no handoff document at all!
2. Why is the CLI command introduced in this PR called `stash stash handoff` rather than `stash handoff`? I assume that's a mistake, why not just call it `stash handoff`?
3. Relatedly, the CLI command introduced in this PR is used to *receive* a stash rather to *give a stash to another agent*, right? Why not call it `stash brief` or something similar?
4. Nit: For the CLI command, instead of outputting the handoff document to stdout, why not save it to `.stash-handoff.md` or something like that? Then the agent can interact with it in a progressive-disclosure-y way if it wants, and also it's persistent. Also something about using stdout as a load-bearing connector between our agent and an API just feels jank.
5. Can users tell that, once they edit a stash summary, it becomes pinned (which means it cannot be edited by the curator agent anymore)? We should have a clear UI affordance showing that editing a Stash summary pins it.
6. Naming nit: what if we called the `curator` agent the `handoff_writer` agent, since that's what it does right now. I agree that we'll want it to grow into the role of being a more general llm wiki maintainer in the future.
7. Why have the AI rewrite SOPs every single time it rewrites the Handoff using `HANDOFF_PRINCIPLES.md`? Is the idea that we don't want to risk losing some ground truth to slop? 
8. can we rename `handoff_summary` -> `handoff_metadata`?
9. can we rename `sessions.status` -> `sessions.summary_status` with enum type `need_summary | in_progress | failed | done`?
10. Can we have the *backend* worker flip status to `summarizing` rather than the user's coding agent?

## Comments about our system more broadly 
1. As we discussed earlier today, we should have separate tables for `stashes` and `workspaces`. Furthermore, we should get rid of the `/api/v1/stashes` endpoint since it's duplicated by `/api/v1/workspaces/{workspace_id}/stashes`
2. Why do we build the agent harness from scratch in ask-the-stash and the handoff curator? Why not just use the claude agent sdk for (1) ask the stash (2) curator (3) session summaries and then do chat titles with a super small completions api call?
3. can we rename /spine -> /overview? Spine is such a weird word and keeps confusing me. Eg
`/api/v1/stashes/{stash_id}/spine` -> `/api/v1/stashes/{stash_id}/overview`

henry-dowling (changes requested):
Okay, here are a lot of thoughts inspired by reading this PR. Some of these are thoughts that don't directly apply to the PR but rather just occurred to me as I began to understand our system more deeply while reviewing this PR:

## Directly relevant to this PR:

1. Why have a forced handoff regeneration path? It should not be on the caller to know the internal freshness state of the handoff. We should just promise them that they'll get a good handoff document every time they ask for one. I don't really see the utility of ever giving them an out-of-date handoff document- that feels worse than no handoff document at all. So in practice this means we need to find some way to keep the handoff up-to-date. We could try regenerating it constantly, or, if that's too costly, I'd advocate for just not making it available when it's out-of-date.
- let's just have two commands, one to regenerate the handoff and one to read the handoff


3. Why is the CLI command introduced in this PR called `stash stash handoff` rather than `stash handoff`? I assume that's a mistake, we should just call it `stash handoff write`

5. Relatedly, the CLI command introduced in this PR is used to *receive* a stash rather to *give a stash to another agent*, right? Why not call it `stash handoff read` or something similar? TODO whatever we choose should be consistent with how we use verbs in the cli

7. Can users tell that, once they edit a stash summary, it becomes pinned (which means it cannot be edited by the curator agent anymore)?

9. Naming nit: what if we called the `curator` agent the `handoff_writer` agent, since that's what it does right now. I agree that we'll want it to grow into the role of being a more general llm wiki maintainer in the future.

11. What's the reasoning for including `HANDOFF_PRINCIPLES.md` verbatim in the handoff? Would it be correct to call it something like `HANDOFF_HUMAN_SECTION`?
- Place it with a new field that's like "Stash Description", which is a human-written description saying what is in the stash. The AI should be able to see that when it's generating its curated summary. 

13. can we rename `handoff_summary` -> `handoff_metadata`?
14. can we rename `sessions.status` -> `sessions.summary_status` with enum type `need_summary | in_progress | failed | done`?

16. Can we have the *backend* worker flip session status to `summarizing` rather than the user's coding agent?
- **Henry can make this change**
 


## Comments about our system more broadly 
1. As we discussed earlier today, we should have separate tables for `stashes` and `workspaces`. Furthermore, we should get rid of the `/api/v1/stashes` endpoint since it's duplicated by `/api/v1/workspaces/{workspace_id}/stashes`
2. Why do we build the agent harness from scratch in ask-the-stash and the handoff curator? Why not just use the claude agent sdk for (1) ask the stash (2) curator (3) session summaries and then do chat titles with a super small completions api call?
        - let's use claude agent sdk for all three
4. can we rename /spine -> /overview? Spine is such a weird word and keeps confusing me. Eg
`/api/v1/stashes/{stash_id}/spine` -> `/api/v1/stashes/{stash_id}/overview`
- what if we had a separate sidebar endpoint so that we can cache it?

henry-dowling:
Also I noticed that generating a summary via the UI has a bug

https://www.loom.com/share/9ff4f8e22ff8462cbced1b68b66d99ec?from_recorder=1&focus_title=1

Codex believes the cause is that we are passing the wrong type of integer into a function (I didn't verify)

from codex:
```
hashtextextended() returns a 64-bit bigint, but the two-argument advisory lock form requires 32-bit ints. For
  some workspace IDs, the cast to int throws integer out of range, so GET /handoff?wait=true returns 500 before
  regeneration starts.

  I hit this locally by clicking “Generate now” on an empty stash. The handoff row stayed stale with no
  last_attempt_at / last_error, because the failure happened before handoff_curator.regenerate().

  Could we switch to the single-argument bigint advisory lock form and include the namespace in the hashed string,
  e.g.:

  pg_try_advisory_lock(hashtextextended($1 || ':' || $2, 0))

  and similarly for unlock?
```

PR #326, opened 2026-05-18
Fix page downloads and markdown editor sync

samzliu:
## Summary
- Fix markdown/html download options so markdown pages export .md + .pdf and HTML pages export .html + .pdf.
- Replace browser print PDF export with generated text/vector PDFs.
- Remove duplicated markdown title/updated block and the unsupported command hint.
- Prevent stale page save responses from resetting in-progress markdown edits.
- Support pasted/dropped image uploads in markdown and make rendered h3 headings more distinct.
- Use the shared download menu for table CSV export and file downloads.

## Verification
- npm run lint
- npm test
- npm run build

## Screenshots
- Captured locally with mocked page data: /tmp/moltchat-downloads-markdown.png
- Not uploaded to the PR because public screenshot upload was blocked by approval policy.

### DIFF
--- a/backend/services/storage_service.py
+++ b/backend/services/storage_service.py
@@ -7,6 +7,7 @@ Configured via environment variables.
 import logging
 import os
 from datetime import UTC
+from urllib.parse import quote
 from uuid import uuid4
 
 import httpx
@@ -40,6 +41,10 @@ def _storage_key(workspace_id: str | None, filename: str) -> str:
     return f"{prefix}/{uuid4().hex[:12]}/{filename}"
 
 
+def _object_uri(key: str) -> str:
+    return f"/{S3_BUCKET}/{quote(key, safe='/')}"
+
+
 async def upload_file(
     workspace_id: str | None,
     filename: str,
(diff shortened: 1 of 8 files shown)


samzliu:
Follow-up from local testing: image upload failures were caused by S3 path signing for filenames with spaces/reserved characters, e.g. screenshot names. Added commit 095c7f7 to URL-encode object paths consistently for PUT/GET/DELETE and added storage regression tests. Manual local API upload of Must be connected to a terminal.
 now returns 201.

samzliu:
Closing in favor of a fresh PR off current main. Main has moved significantly in the last day (FileViewerHeader extraction, page comments, skeleton states, descriptions-as-editor) and touches the same files, so reapplying the intent on top of current main is cleaner than resolving conflicts.

Intent being re-shipped:
- URL-encode storage keys in storage_service.py (uploads with spaces / # in filenames)
- Generated PDF (text/vector) for page downloads instead of window.print()
- HTML vs markdown pages get different download options
- Stale save-response guard on the page so older save replies don't clobber newer state
- Paste / drop image uploads in the markdown editor
- Bigger markdown heading sizes
- Shared download menu for table CSV export (cleaner auth via fetchAuthed)

PR #328, opened 2026-05-18
Fix mismatched-transaction crash when uploading images into a markdown page

samzliu:
## Summary

Dropping or pasting an image into the markdown editor crashed with **\"Applying a mismatched transaction\"** at:

\`\`\`tsx
editor.chain().focus().setImage({ src: result.url, alt: result.name }).run();
\`\`\`

### Why
During the \`await uploadFile(...)\` between drop and dispatch, the \`resolvedMarkdown\` effect higher in this file can fire \`editor.commands.setContent(initialContent)\` to swap in URL-resolved content (rewriting relative \`\` refs to absolute signed URLs after \`listFiles\` resolves). The chain then dispatches against the new editor state, but TipTap's chain captured the pre-rehydrate position — ProseMirror's \`view.updateState\` rejects the transaction because its \`before\` doc no longer matches.

### Fix
Switch from chain to direct \`editor.commands.X\` calls. Each command builds its own transaction from the current state at dispatch time, so there's no stale-state window. Also bail cleanly if the editor was destroyed mid-upload.

While here: the non-image branch did \`setLink({ href }).insertContent(name)\` which set the link mark on the **prior selection** and then inserted plain text — not what was intended. Replaced with a single \`insertContent\` of a linked text node so the file name itself becomes the link.

## Test plan
- [ ] Open a markdown page, drag in an image → image renders inline; no console error.
- [ ] Paste an image from clipboard → same.
- [ ] Drag in a non-image file (e.g. a PDF) → a link with the file name is inserted; clicking it opens the file.
- [ ] Drop multiple files → all of them insert without error.

### DIFF
--- a/backend/services/storage_service.py
+++ b/backend/services/storage_service.py
@@ -7,6 +7,7 @@ Configured via environment variables.
 import logging
 import os
 from datetime import UTC
+from urllib.parse import quote
 from uuid import uuid4
 
 import httpx
@@ -40,6 +41,10 @@ def _storage_key(workspace_id: str | None, filename: str) -> str:
     return f"{prefix}/{uuid4().hex[:12]}/{filename}"
 
 
+def _object_uri(key: str) -> str:
+    return f"/{S3_BUCKET}/{quote(key, safe='/')}"
+
+
 async def upload_file(
     workspace_id: str | None,
     filename: str,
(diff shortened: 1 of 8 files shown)


samzliu:
Closing in favor of a fresh PR off current main alongside #326's intent. The chain → commands fix for image drop / paste will be re-shipped, since the underlying race (resolvedMarkdown effect rehydrating during the upload await) still applies on top of main's editor.

PR #340, opened 2026-05-19
[codex] Fix description editor focus stealing

henry-dowling:
## Summary
- Prevent the shared description editor frame from refocusing Tiptap when a click already landed inside the editable ProseMirror content.
- Keep frame/background clicks focusing the editor for the intended large-click-target behavior.
- Add focused regression coverage for the click-target guard.

## Root cause
The editor frame handled every bubbled click by calling `editor.commands.focus()`. During rapid blur/refocus cycles, that focus command could restore a stale editor selection after ProseMirror had placed the cursor at the clicked text position, which made the cursor jump away from the click target.

## Validation
- `cd frontend && npm test`
- `cd frontend && npm run lint`
- `cd frontend && npm run build`
- Headless Playwright check against the local workspace page: repeated blur/click cycles kept the selection in the clicked paragraph at offset `0` with no console errors.

Product screenshot attached in the PR thread.

### DIFF
--- a/frontend/src/components/DescriptionEditor.tsx
+++ b/frontend/src/components/DescriptionEditor.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useCallback, useEffect, useRef } from "react";
+import { useCallback, useEffect, useRef, type MouseEvent } from "react";
 import { EditorContent, useEditor } from "@tiptap/react";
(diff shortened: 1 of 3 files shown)


henry-dowling:
Product screenshot: workspace description editor after the focus fix.

PR #406, opened 2026-05-21
feat(slides/exports): 2x DPI + invisible text overlay + shared canvas dims

samzliu:
## Summary

PR 3 of 3 from the slides-improvements plan. Turns the PPTX / Google Slides exports from \"screenshots of slides\" into artifacts users can actually select, copy, and search inside.

- **`backend/exports/constants.py`** — single source of truth for the slide canvas (1920×1080 px, 13.333\"×7.5\", 16:9, 2× device scale). All three exporters (PPTX, PDF, Google Slides) now derive their dimensions from it. Magic numbers eliminated.
- **2× device scale** on PPTX screenshots — the embedded PNGs are now 3840×2160 instead of 1920×1080. Slides stay crisp when zoomed in PowerPoint / projected on a 4K screen, at ~4× file size.
- **Invisible text overlay** in PPTX — for each slide we harvest text from `<h1>…<h6>, <p>, <li>, <td>, <th>, <blockquote>, <figcaption>` with bounding boxes, and emit transparent text shapes at matching EMU coords. Text is now selectable, copyable, and searchable in PowerPoint / Keynote / Slides without reconstructing native styled shapes (which the code comment had previously called \"a rabbit hole\").
- **PDF tightening**: `html, body { overflow: hidden }` in the print CSS so a slide that overflows the 16:9 page gets clipped at the page boundary instead of leaking visual artifacts onto adjacent pages.

The Google Slides exporter (`backend/integrations/google/exporters/slides.py`) inherits the upgrade for free — it just calls `build_pptx_bytes_for_page` and uploads the result.

## Why now

Reported: \"current exports have fidelity loss across pptx, slides, pdf.\" The image-embed approach was deliberately simple but cost users the ability to do anything other than view the slides. The text-overlay trick gets the most-asked-for property (selection + search) without committing to a long native-shape conversion project.

## Test plan

- [x] `tests/test_export_constants.py` (new) — locks 1920 / 1080 / 16:9 / 2× and grep-asserts that pptx.py and pdf.py don't re-introduce hardcoded dims.
- [x] Smoke-test: `_capture_all_slides()` returns correct text blocks with bounds for a 2-slide HTML; generated PPTX contains `<a:alpha val=\"0\"/>` confirming the text overlay is invisible; `_render_pdf()` produces a valid 24 KB PDF (vector text, 16:9 page size). Screenshot of the rendered PDF page 1 attached.
[...]

PR #412, opened 2026-05-21
fix(slides/exports): launch chromium with --no-sandbox + --disable-dev-shm-usage

samzliu:
## Summary

Fixes the PPTX export hang on prod that surfaced during QA after #406 shipped.

After #406 the PPTX path captures each slide at \`device_scale_factor=2\` (3840×2160 PNGs) via Playwright. On the Render celery worker that crashes Chromium hard enough that Render's process supervisor declares the instance dead and restarts it ~22 s into a 5-slide deck. The celery task disappears mid-run and the result stays \`PENDING\` forever.

Two Chromium launch args fix it:

- \`--no-sandbox\` — the worker container runs as a non-root user without the capabilities Chromium's sandbox needs.
- \`--disable-dev-shm-usage\` — Render containers ship with a 64 MB \`/dev/shm\`; the 2× DPI shots blow past that. Falling back to \`/tmp\` avoids the crash.

PDF export currently works (one \`page.pdf()\` call, lighter on shm). Applied the same args there for consistency so it doesn't regress when render paths change.

## Why this slipped local QA

I exercised the export pipeline directly via the venv on my Mac, not in a docker container. Both \`/dev/shm\` and the sandbox are non-issues there. The first real container run was on Render this morning, which is when it surfaced.

## Test plan

- [x] \`black --check\` + \`ruff check\` clean.
- [ ] After merge + deploy: trigger a PPTX export against the QA deck (\`https://app.joinstash.ai/workspaces/7553df78-bf9a-445e-b99c-bc3316113b4a/p/7d9f8bd5-ca7b-4d67-8360-3d78682839a0\`) and confirm the task transitions \`PENDING → STARTED → SUCCESS\` with a downloadable URL.
- [ ] Worker doesn't restart during the export (check Render metrics + logs).
- [ ] Spot-check the resulting PPTX has the expected 3840×2160 PNGs and \`<a:alpha val="0"/>\` text overlays.

## Out of scope

The R2 bucket (\`boozle\`) CORS policy doesn't allow \`https://app.joinstash.ai\` as a \`GET\` origin, so even successful exports get blocked in the browser. Tracked separately — needs a Cloudflare config change, not code.

### DIFF
--- a/backend/exports/pdf.py
+++ b/backend/exports/pdf.py
@@ -61,7 +61,10 @@ def _inject_paged_css(html: str) -> str:
 
 async def _render_pdf(html: str) -> bytes:
     async with async_playwright() as p:
-        browser = await p.chromium.launch()
+        # See pptx.py for why these args matter in the Render worker.
+        browser = await p.chromium.launch(
+            args=["--no-sandbox", "--disable-dev-shm-usage"],
+        )
         try:
             page = await browser.new_page(
                 viewport={"width": SLIDE_WIDTH_PX, "height": SLIDE_HEIGHT_PX},
--- a/backend/exports/pptx.py
+++ b/backend/exports/pptx.py
@@ -1,14 +1,19 @@
-"""PPTX export — rasterized slides + invisible text overlay.
+"""PPTX export — rasterized slides.
 
 Uses Playwright to screenshot each `<section class="slide">` as a PNG
 at 2x device scale, then assembles a 16:9 deck with one image per
-slide and a transparent text layer harvested from the rendered DOM.
-The text layer makes the exported PPTX selectable, copyable, and
-searchable in PowerPoint / Keynote / Google Slides, without trying to
-reconstruct native shapes (which would be a rabbit hole).
+slide.
 
 Reused by the Google Slides exporter (uploads the same PPTX to Drive
 with `convertTo=true`).
+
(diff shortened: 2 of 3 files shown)


samzliu:
Pushed two more fixes after QA on prod showed broken render on slide 2:

**Bug A — slide section didn't fill the canvas in the screenshot.** PR #406's pptx.py didn't inject the canvas-enforcing CSS the in-app viewer (`injectSlideDeckBootstrap` in HtmlPageView.tsx) injects. Agent slides without explicit dimensions captured at content-height, ending up as a thin band on a white page in PowerPoint. `_build_single_slide_html` now mirrors the viewer's CSS so screenshots always fill 1920×1080.

**Bug B — 'invisible' text overlay wasn't invisible in PowerPoint.** `<a:alpha val="0"/>` inside `<a:srgbClr>` on a text run is technically valid OOXML, but PowerPoint's text-run renderer ignores it. Every label rendered as visible white text on top of the slide image — doubling every title and body. I dropped the overlay entirely; accept the loss of selectable PPTX text as the lesser evil for now. The module docstring explains so we don't reintroduce without a new approach.

Local smoke test: rebuilt the well-formed-slide screenshot. Purple now fills the entire 3840×2160 canvas; text appears once.

PR #506, opened 2026-06-05
Add api_key auth path + Gong as an agent source

samzliu:
## What

**PR 2 of 3.** Adds a small, reusable **`api_key` auth path** for providers that authenticate with pasted credentials instead of an OAuth redirect, then adds **Gong** on top of it.

> Stacked on #505 (Jira/Asana). Base is `integrations-jira-asana`; review/merge that first. GitHub will retarget this to `main` once #505 merges.

## Why api_key

Gong OAuth requires Gong to whitelist our app as a marketplace partner (external dependency). Per-customer **Access Key + Secret** (HTTP Basic) needs no approval — but the framework was OAuth-redirect-only. This adds the missing path, reusable for Snowflake in PR 3.

## How

**Framework** (`base.py` + `router.py`):
- `CredentialField` + the api_key contract: a provider sets `auth_kind="api_key"`, declares `credential_fields`, and implements `connect_with_credentials`, which validates the creds upstream and returns them JSON-encoded as the access token (no expiry) so they ride the **same encrypted storage column** as OAuth tokens. No DB schema change, no token-storage change.
- `POST /integrations/{provider}/credentials` validates + stores; the providers list now surfaces `auth_kind` + `credential_fields` so the UI can render a form. Bad/rejected credentials surface as **400**, not 500.

**Gong**:
- Access Key + Secret validated against `/v2/workspaces`. One connection = one source (`gong_calls`). Calls + transcripts from a rolling 90-day window become **searchable** documents (title, date, speaker-labelled transcript) in `gong_documents` (migration 0089).

**Frontend**: inline credential form (password fields built from `credential_fields`) revealed when an api_key provider's Connect is clicked; Gong connector card + icon.

## Verification

- `ruff` + `pytest` (`test_integration_sources.py`: transcript render, api_key wiring, missing-credential rejection) — green. Full `test_sources.py` suite (26 tests) green.
- Frontend `tsc --noEmit` + `eslint` — green. Migration 0089 applies; `gong_documents` matches the `github_documents` shape.
- Throwaway backend confirmed: Gong lists `enabled` with `auth_kind=api_key` and two secret credential fields; `POST /gong/credentials` with bad keys returns **400** (validated against the live Gong API, which 401s); `add_source` for `gong_calls` without a connection returns 401.
- UI: the Gong connector card renders in Settings → Sources (screenshot in thread). The credential form (verified by tests/API) shows on Connect once the backend reports `auth_kind=api_key`.

Connecting for real needs a customer's Gong Access Key + Secret (with `api:calls:read:*` scopes) — entered in the form, no server-side env vars.

### DIFF
--- a/backend/integrations/__init__.py
+++ b/backend/integrations/__init__.py
@@ -17,6 +17,7 @@ for a new provider — only the explicit list of providers grows.
 from . import (
     asana,  # noqa: F401
     github,  # noqa: F401 — import for side-effect (provider self-registration)
+    gong,  # noqa: F401
     google,  # noqa: F401
     granola,  # noqa: F401
     jira,  # noqa: F401
--- a/backend/integrations/base.py
+++ b/backend/integrations/base.py
@@ -30,6 +30,16 @@ class AccountInfo:
     display_name: str | None
 
 
+@dataclass
+class CredentialField:
+    """One input in an api_key provider's connect form (see below)."""
+
+    name: str
+    label: str
+    secret: bool = False  # render as a password field + never echo back
+    placeholder: str = ""
+
+
 class Integration(Protocol):
     name: str
     """URL segment + provider key. e.g. 'google', 'github'."""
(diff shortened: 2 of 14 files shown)


samzliu:
Landed on `main` as part of #507 — its commit `51e2553e` (api_key auth + Gong) is in main's history. GitHub auto-closed this PR when its stacked base branch was deleted during the merge sequence, so it shows closed rather than merged, but the code is fully merged.

PR #524, opened 2026-06-05
Harden auth and export boundaries for managed data

henry-dowling:
## Summary
- make workspace role checks write-aware so viewers cannot mutate pages, files, sessions, tables, or collab documents through `permission_service`
- restrict sharing and workspace/public Stash publication to workspace owners/admin paths
- block unauthorized one-call `/publish` attempts before creating pages, so viewers cannot create private drafts and non-owner editors cannot leave orphaned pages while attempting public/workspace Stashes
- enforce read-only workspace roles across session ingress: viewers can no longer push live events, batch events, upload transcripts, create session shells, upload artifacts, or materialize sessions into pages
- remove self-hosted workspace UUID auto-join: non-members now get 404 instead of being silently added as editors when requesting `/workspaces/{id}`
- fix session-folder assignment to verify writer role plus session/folder workspace boundaries
- restrict session-folder management to the folder owner or workspace owner; explicit write shares can write contents but cannot rename/delete the folder itself
- remove public-write Stash and session-folder links; public links are read-only at API validation, service permission checks, and database constraints, with existing public-write rows downgraded to read
- require editor/owner workspace roles for workspace-wide Stash write access, so workspace-write Stashes no longer let viewers create shared pages
- harden exports by rechecking worker-side page access, blocking Playwright outbound network requests, and only resolving authorized Stash file images
- require task status polling to match a server-side task ownership record for exports and source sync jobs
- add per-source settings and restrict Slack ingestion/search/read/counts to an explicit `allowed_channel_ids` allowlist, with no channels copied by default
- expose Slack channel allowlist entry in the integration UI so Slack sources cannot be added without explicit channel IDs
- require Gong sources to declare explicit `allowed_workspace_ids`; scheduled sync and historical backfill copy no calls by default and ignore calls outside the allowlist
- persist Gong workspace IDs on cached calls and enforce the current allowlist on list/read/search/count, so tightening a Gong source cannot leave older cached calls visible
- expose Gong workspace allowlist entry in the integration UI so Gong sources cannot be added without explicit workspace IDs
- validate Jira source refs server-side and quote Jira JQL project clauses, blocking direct API calls from widening a Jira source with malformed `{cloudId}:{projectKey}` values
- enforce read-only workspace roles across connected-source mutations: viewers can list/read/query their allowed sources but cannot add sources, trigger history backfills, enqueue manual syncs, or delete source rows
[...]

PR #528, opened 2026-06-08
[security split 1/10] Core auth, export, and task boundaries

henry-dowling:
Closes the first set of backend security holes from the demo hardening pass: server-side request forgery in exports, an IDOR on async task status, unscoped Slack ingestion, and missing role checks on sharing and write paths.

## Export pipeline can no longer reach the network

Exports previously fetched arbitrary `http(s)` image URLs server-side and rendered pages in a Playwright browser with full network access — an SSRF vector into anything the backend can reach.

- `ImageFetcher` drops support for remote URLs entirely. Images in exported decks are limited to inline data URIs (now subject to the 25 MB cap too) and same-workspace Stash files.
- Stash file resolution is defense-in-depth: the workspace ID in the URL, the file row's actual workspace, and a `permission_service.check_access` call for the exporting user must all agree.
- Every Playwright context (PDF, raster, layout probe) aborts all network requests via `route("**/*")`. Inline data URIs still render; remote images intentionally go blank.
- PDF/PPTX export now verifies the requesting user can read the page before rendering (PPTX export plumbs `user_id` through, including the Google Slides path).

## Task status is owner-gated (IDOR fix)

`GET /api/v1/tasks/{task_id}` let any authenticated user poll any Celery task ID — and export results include download URLs. A new `task_records` table (migration `0100`) records the owner at dispatch time; both endpoints that hand task IDs to clients (page export, source sync) register them, and status polls from anyone else 404.

## Slack sources are channel-allowlisted

A connected Slack source used to ingest every channel the token could see. Sources now carry a `settings.allowed_channel_ids` allowlist (migration `0101`), enforced at every layer: backfill, Events-API webhook ingest, on-demand history fetch, list/read/count, and FTS search. An empty allowlist fails closed — the source ingests and serves nothing.

## Workspace roles actually mean something

- `check_access` now distinguishes roles: owners/editors can write, viewers are read-only (previously any workspace member had full access). Wired to main's tiered `require` levels (read < comment < write).
- Creating shares, and creating/publishing workspace-visible or public Stashes, is restricted to workspace owners.
- Session-folder create/assign requires write; assignment validates that both session and folder belong to the workspace and the caller can write both (cross-workspace assignment was previously possible).

## Test hardening

`conftest` truncates every table (instead of a hand-maintained list with swallowed errors), so new tables can't silently leak state between tests.

---

**Logistics:** split from #524, slice 1/10. Merge the stack in order 1→10; each later PR bases on the previous branch so its Files tab stays scoped. `origin/main` has been merged in (migrations renumbered to `0100`/`0101` to avoid Alembic ID collisions with main's `0093`–`0099`; role gate adapted to main's tiered `require` API).

**Validation** on the merged result: full backend suite 303 passed / 3 skipped against a fresh database (exercises the new migration chain), `ruff check` clean.


henry-dowling:
Resolved the merge conflict with main (merged `origin/main` in rather than rebasing, to keep the 2/10–10/10 stack branches valid). The resolution went beyond the one textual conflict in `test_sources.py`:

- **Renumbered migrations** `0093_task_records` → `0100` and `0094_source_settings` → `0101` (parented on `0099`) — main had since merged its own `0093`–`0099`, which would have produced duplicate Alembic revision IDs.
- **Adapted the `check_access` role gate to main's tiered `require` API** (read < comment < write, from #559's share-comment-tier work): owner/editor pass any tier, viewer passes only `read`. Converted the `require_write=True` call sites in `session_folder_service` and tests to `require="write"` — the auto-merge would otherwise have produced a `NameError`.

Validation on the merged result: full backend suite 303 passed / 3 skipped against a fresh DB (which also exercises the renumbered migration chain via `alembic upgrade head`), `ruff check` clean.

Note for the stack: branches 2/10–10/10 still base on the old tip of this branch and will need this merge cascaded into them before they can land.

PR #557, opened 2026-06-09
Fix stale table workspace links

henry-dowling:
## Summary

- Recover stale `/tables/[tableId]?workspaceId=...` links by checking the authenticated cross-workspace table index after the scoped lookup fails.
- Redirect the URL to the table's canonical workspace and load rows from that workspace.
- Add a regression for stale workspace links resolving to the canonical table workspace.

## Root Cause

The table page trusted the `workspaceId` query param. If that workspace was stale or no longer accessible, the first scoped table lookup failed and the UI showed `Table not found`, even when the same table was still visible to the user in another workspace.

## Product Impact

Old table links now self-heal instead of looking like data loss. This covers the `Hiring Outreach CRM` failure mode we just debugged.

## Screenshots

No screenshot attached: this is a route-resolution fix with no visual UI change. The user-facing result is the existing table page loading instead of the existing `Table not found` error.

## Validation

- `npm test -- --run 'src/app/tables/[tableId]/page.test.tsx'`
- `npx eslint 'src/app/tables/[tableId]/TableClient.tsx' 'src/app/tables/[tableId]/page.test.tsx'`
- `git diff --check -- 'frontend/src/app/tables/[tableId]/TableClient.tsx' 'frontend/src/app/tables/[tableId]/page.test.tsx'`

### DIFF
--- a/frontend/src/app/tables/[tableId]/TableClient.tsx
+++ b/frontend/src/app/tables/[tableId]/TableClient.tsx
@@ -423,8 +423,21 @@ function TableEditorPageInner() {
         return;
       }
       if (resolvedWorkspaceId) {
-        setTable(await getTable(resolvedWorkspaceId, tableId));
-        return;
+        try {
+          setTable(await getTable(resolvedWorkspaceId, tableId));
+          return;
+        } catch {
+          const all = await listAllTables();
+          const match = all?.tables?.find((t) => t.id === tableId);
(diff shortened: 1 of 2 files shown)


henry-dowling:
Superseded by #561: instead of recovering stale workspaceId links client-side, table URLs no longer carry the workspace at all — the backend resolves it from the table row, so links can't go stale.

PR #564, opened 2026-06-10
dev: document uv-based local CLI development

henry-dowling:
## What

Documents the local-CLI dev workflow and standardizes all Python installs on uv:

- By default `stash` is the released PyPI build (`uv tool` install, self-updating via the plugin's session-start hook).
- To develop the CLI in a checkout (main repo or any worktree): `uv venv -p 3.12 && uv pip install -e .` once, then `source .venv/bin/activate` per terminal. The `(.venv)` prompt badge shows dev mode is on; new terminals default back to the released CLI.
- Backend dep install now targets `.venv` via `uv pip` instead of bare `pip`.

## Why

A bare `pip install -e .` under a pyenv interpreter is exactly what left the stale `stash` binary on PATH that #560 had to defend against. This makes the venv path the documented default with the simplest possible mental model: activate = this checkout's code, otherwise = released CLI.

(Earlier revisions of this PR shipped a committed `.envrc` + direnv for automatic per-directory switching; dropped deliberately — one more tool and invisible PATH state wasn't worth it vs `source .venv/bin/activate`.)

## Verification

- `source .venv/bin/activate` → `command -v stash` → `.venv/bin/stash`, `import cli` resolves to the working tree
- `deactivate` → `command -v stash` → `~/.local/bin/stash` (uv release)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,8 +86,6 @@ By default `stash` is the released PyPI build (`uv tool` install, self-updating)
 1. Once per checkout: `uv venv -p 3.12 && uv pip install -e .`
 2. Per terminal: `source .venv/bin/activate` — your prompt shows `(.venv)` and `stash` now runs this checkout's working tree. New terminals default back to the released CLI.
 
-Use uv for all Python installs — never a bare interpreter's pip (a bare `pip install -e .` under pyenv leaves a stale `stash` binary on PATH that shadows the real install).
-
 ### Backend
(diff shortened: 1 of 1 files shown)


henry-dowling:
Dropped the "never a bare interpreter's pip" admonition from this PR — that hazard is now detected mechanically instead of remembered: #565 warns at session start and after `stash login` whenever more than one `stash` install is on PATH. The doc here stays purely as the happy-path workflow.

PR #619, opened 2026-06-13
Pages: comment replies + delete buttons + names (also restores #618, wiped from main)

samzliu:
## ⚠️ This also repairs main
**#618 (the full comment edit/delete CRUD) merged but was then clobbered off `main` by a concurrent push to the shared checkout** — the classic shared-checkout race. The result: `main` has the comment *create* code but none of #618's edit/delete endpoints, www comment routes, proxy rewrites, agent docs, **or the `0113` migration** — yet prod's DB is stamped at `0113`. So **the next prod deploy from main would fail** (`KeyError: '0113'`, which I hit locally). This PR restores all of #618 and makes its migration idempotent (`IF NOT EXISTS`) so it's safe on prod, which already ran the original.

Verified the repaired chain on a fresh DB: `0112 → 0113 → 0114` ends with `paste_comments` having both `edit_token` and `parent_id`.

## New in this PR (the actual ask)
**Delete buttons:**
- **Edit page**: a Delete on every comment, using the page's edit token → owner moderation (works after reload).
- **View page**: a Delete only on comments you posted *this session* (the one-time comment token), so you can remove your own — but not others'.

**Threaded replies (like the main app):**
- `parent_id` on comments (migration 0114). A reply hangs off a top-level comment; deleting a comment cascades to its replies.
- Rail nests replies under their parent with a **Reply** composer. Agents can reply too: `POST …/comments` with `parent_id`.

**Commenter names:**
- The composer's name field is remembered in `localStorage`, so it prefills across comments.

## Authorization recap (who can delete a comment)
Comment author (one-time comment token) **or** page owner (page edit token, for moderation). Empty/wrong token → 404.

Verified end to end in-browser: nested reply renders, reply composer posts with parent_id, owner-delete on edit page, author-only delete on view page, name persists. 34 backend tests; www build + lint clean.

### DIFF
--- a/backend/migrations/versions/0113_paste_comment_tokens.py
+++ b/backend/migrations/versions/0113_paste_comment_tokens.py
@@ -1,0 +1,34 @@
+"""Per-comment edit tokens for the pastes pastebin.
+
+Restored: this migration was lost from main (a concurrent push to the
+shared checkout swept the file) while the code that depends on the
+column stayed — leaving the chain broken and prod stamped at a revision
+with no file. Re-adding it repairs the chain. ``IF NOT EXISTS`` keeps it
+safe on databases that already ran the original (prod).
+
+Comments are anonymous like the pages they hang on. To let the author
+edit or delete a comment later — and let agents manage comments they
+post — each comment gets its own write token, minted at create time and
+returned once (same model as a paste's edit_token). Deletion also
+accepts the parent paste's edit_token so the page owner can moderate.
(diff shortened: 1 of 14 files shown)


samzliu:
Superseded by a clean rebase: #618 was re-merged to main independently, so this PR's #618-restore became redundant and conflicting. Reopening the net-new reply/delete/names work on a fresh branch off main.

PR #629, opened 2026-06-15
Share dialog: change a person's permission after sharing

henry-dowling:
## What

You can now change someone's permission **after** you've shared with them — no more remove-and-re-invite. Each row under **People with access** has a dropdown to switch between **Can view / Can comment / Can edit**.

Closes the request: "share with someone to view and then change it to comment."

## How

- `ResourceShareButton` renders a permission `<select>` per share row (replacing the static label). Changing it calls the existing share endpoint, which already upserts `ON CONFLICT DO UPDATE SET permission`, so re-sharing the same email with a new permission just updates the existing share or pending invite.
- Added a **Can comment** option — the backend already accepts `read`/`comment`/`write`; the frontend just wasn't offering it. Widened `GeneralPermission` to include `comment`.
- The owner row stays static ("Owner"). Pending invites are editable too (the invite table upserts the same way).

## Tests

- New unit test: changing a person's permission calls `shareObjectByEmail(..., "comment")` and shows "Access updated." All 3 `ResourceShareButton` tests pass.

## Screenshot

Sam switched from "Can view" to "Can comment" in the live component:

### DIFF
--- a/frontend/src/components/share/ResourceShareButton.test.tsx
+++ b/frontend/src/components/share/ResourceShareButton.test.tsx
@@ -123,4 +123,33 @@ describe("ResourceShareButton", () => {
     expect(await screen.findByText("ada@example.com")).toBeInTheDocument();
     expect(screen.getByText("Invited")).toBeInTheDocument();
   });
+
+  it("changes an existing person's permission", async () => {
+    render(
+      <ResourceShareButton
+        objectType="page"
+        objectId="page-1"
+        resourceName="Blog post outline"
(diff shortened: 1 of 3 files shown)


henry-dowling:
### Screenshot — changing Sam from "Can view" → "Can comment"

Rendered against the real `ResourceShareButton` component; "Access updated." confirms the change persisted.

Share dialog with per-person permission dropdown set to Can comment

Durable copy (Stash): https://app.joinstash.ai/f/bb26328f-a883-4512-b410-89ebd2ce6743

PR #634, opened 2026-06-15
Stop new tables from inflating the select and # columns

henry-dowling:
## Problem

When you create a new (empty) table, the left checkbox column and the `#` row-number column render absurdly wide.

## Root cause

The grid is `table-fixed w-full`. The trailing row-actions column took a fixed width from its header cell's `w-16` class. With **0 data columns**, that left *no* auto-width column to soak up the leftover horizontal space, so the browser spread the slack proportionally across every column — fattening the checkbox and `#` columns.

Measured in an isolated repro of the exact markup at a 1900px container:

| Column | Before (`w-16` spacer) | After (auto spacer) |
|---|---|---|
| checkbox | **315px** | **32px** |
| `#` | **394px** | **40px** |
| `+` add | **394px** | **40px** |
| trailing spacer | 797px | 1788px |

## Fix

Drop the fixed width on the trailing header cell so it stays the auto spacer. The fixed columns keep their real widths and the spacer absorbs the remaining width.

## Verification

Before/after of the exact table markup (empty table). Top = before, bottom = after:


henry-dowling:
Before (top, fat columns) vs after (bottom, tight columns), exact table markup at 1900px:

PR #638, opened 2026-06-15
Allow disconnecting a source from the connect modal

henry-dowling:
## What

A connected source in the **Connect a source** modal previously only offered an **Open** button — to disconnect, you had to open the integration's detail page first. This adds a quiet **Disconnect** action right in the modal next to **Open**.

## How

- `SourceConnectorList` now wires the connected-row action to `disconnectIntegration(provider)`, the same backend endpoint (`POST /api/v1/integrations/{provider}/disconnect`) the detail page already uses.
- Disconnect is guarded by the shared `useConfirm` dialog ("Disconnect {provider}? You'll need to reconnect to sync its sources again.") — mirroring the detail-page wording and styling exactly.
- After a successful disconnect the list refreshes, so the row flips back to **Connect**.

No backend changes — the endpoint, token revocation, and source cleanup already exist.

## Tests

New `SourceConnectorList.test.tsx`:
- confirms then disconnects the provider and refreshes the list
- cancelling the confirmation does nothing

`npm run lint` and the new vitest file both pass.

## Verification

Manually verified against the local stack with a connected GitHub source — see the screenshot in the PR thread. The connected row shows **Disconnect**; unconnected providers still show **Connect**.

### DIFF
--- a/frontend/src/components/integrations/SourceConnectorList.test.tsx
+++ b/frontend/src/components/integrations/SourceConnectorList.test.tsx
@@ -1,0 +1,81 @@
+import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { IntegrationStatus } from "../../lib/integrations";
+import { ConfirmDialogProvider } from "../ConfirmDialog";
+import SourceConnectorList from "./SourceConnectorList";
+
+const listIntegrations = vi.fn();
+const disconnectIntegration = vi.fn();
+
+vi.mock("@/lib/integrations", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../lib/integrations")>();
+  return {
+    ...actual,
+    listIntegrations: () => listIntegrations(),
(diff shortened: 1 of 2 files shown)


henry-dowling:
### Manual verification ✅

Logged into the local stack as a test user with a connected GitHub source. The connected source now renders a **Disconnect** action inline (next to **Open** when a workspace is in context), while unconnected providers still show **Connect**:

Disconnect button on a connected source

Clicking **Disconnect** opens the shared confirm dialog ("Disconnect GitHub? You'll need to reconnect to sync its sources again."); confirming calls `POST /api/v1/integrations/github/disconnect` and the row flips back to **Connect**. Both the confirm and cancel paths are covered by `SourceConnectorList.test.tsx`.

henry-dowling:
### Updated: single Connect/Disconnect toggle

Per feedback, the connected row no longer shows two buttons (Open + Disconnect). The action is now a single button that flips from **Connect** to **Disconnect** once the source is connected:

Connect becomes Disconnect when connected

The integration's detail page (for adding specific repos/pages) is still reachable from the sidebar's **External Sources** list. Verified live against the local stack.

PR #648, opened 2026-06-15
Align search popup with the search bar

henry-dowling:
## Problem

When you click the navbar search bar, the command palette popup did not line up horizontally with the bar. The bar is centered within the header's `flex-1` column (between the asymmetric breadcrumb/share sections), while the popup centered itself in the full viewport via `mx-auto` — two different reference frames, so they never aligned.

## Fix

On open, the popup measures the search bar's `getBoundingClientRect()` and positions itself absolutely to match the bar's `left` / `top` / `width` exactly. This makes them line up regardless of sidebar collapse state or breadcrumb width.

## Verification

Measured both elements in the running app — bar and popup dialog are now identical: `left=211, top=8, width=896`. Unit tests pass (`CommandPalette.test.tsx`).

Aligned popup (left edge + width match the bar, sitting flush with the content column):

### DIFF
--- a/frontend/src/components/AppShell.tsx
+++ b/frontend/src/components/AppShell.tsx
@@ -266,6 +266,7 @@ export default function AppShell({
     () => readCachedWorkspaces(user.id)?.all ?? [],
   );
   const [cmdkOpen, setCmdkOpen] = useState(false);
+  const searchBarRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


henry-dowling:
Verified in the running app: the popup dialog and the search bar now have identical bounding boxes — both `left=211, top=8, width=896`. (Screenshot captured locally; GitHub's CLI can't inline-upload images — happy to drop it into the thread via the web UI if useful.)

PR #649, opened 2026-06-15
Ensure push-hook sessions always land in a session folder (FER-219)

henry-dowling:
## Problem

Session rows get created by three different paths, but only `POST /sessions` homed them into a folder. The **transcript-upload** endpoint and the **event-stream** path (`push_event` / `push_events_batch`) called `upsert_session` *without* a folder. So any session whose row was first created by those paths landed at the **workspace root in no folder** instead of a session folder:

- **Every Codex session** — Codex has no session-start hook, so the first streamed event creates the row.
- **Any transcript-created row** — e.g. a CLI that lost connectivity mid-session and only the end-of-session transcript upload reaches the backend.

This is the "land in workspaces and not session folders" bug from FER-219.

## Fix

**Centralize the invariant in `upsert_session`** — the single place every creation path goes through. A new session is born into the workspace **Default** folder when none was pushed. The Default is resolved *only when the row doesn't exist yet*, so an explicit move-to-root (`session_folder_id = NULL`) is never re-homed by a later streamed event (the existing `COALESCE` protects updates).

This let me drop the now-duplicated Default-folder branch from the sessions router.

**Default folder ownership fix:** `ensure_default_folder` now owns the folder as the **workspace owner**, not whoever happened to push the first session. Previously a member who authored the first session "owned" the auto-created Default folder and kept read access to its sessions even after leaving the workspace. (Caught by an existing access-control test once the events path started homing to Default.)

**Honor explicit pins on the transcript path:** thread the manifest's `session_folder_id` through the whole transcript-upload pipeline — plugin `StashClient.upload_transcript` → detached `_do_upload` / `spawn_transcript_upload` (Codex) → `stream_session_end` (Claude) → backend transcript endpoint (validated like the sessions router) — so a pinned repo targets the right folder even when the upload creates the row.

**New CLI command:** `stash sessions use-folder <name|id>` pins this repo's agent sessions to a folder (writing `.stash`), `--default` clears the pin. Previously the only way to set a repo's session folder was to hand-edit `.stash`.

## Tests

- New: an un-targeted Codex-style event-created session lands in the Default folder (not folderless).
- New: a transcript upload with an explicit `session_folder_id` lands in that folder.
- Existing access-control / permission / folder / session suites pass (backend + plugins + CLI).

### DIFF
--- a/backend/models.py
+++ b/backend/models.py
@@ -568,6 +568,9 @@ class HistoryEventCreateRequest(BaseModel):
     event_type: str = Field(..., min_length=1, max_length=64)
     content: str = Field(..., min_length=1)
     session_id: str | None = Field(None, max_length=64)
+    session_folder_id: UUID | None = Field(
+        None, description="Pinned folder for this session (from the repo manifest)"
+    )
     tool_name: str | None = Field(None, max_length=128)
     metadata: dict = Field(default_factory=dict)
     attachments: list[Attachment] | None = None
--- a/backend/routers/memory.py
+++ b/backend/routers/memory.py
@@ -56,6 +56,7 @@ async def push_ws_event(
         content=req.content,
         created_by=current_user["id"],
         session_id=req.session_id,
+        session_folder_id=req.session_folder_id,
         tool_name=req.tool_name,
         metadata=req.metadata,
         attachments=attachments,
--- a/backend/routers/session_folders.py
+++ b/backend/routers/session_folders.py
@@ -94,7 +94,7 @@ async def list_folders(workspace_id: UUID, current_user: dict = Depends(get_curr
     # Non-members may still see folders shared with them, so this isn't gated on
     # membership. Members get a Default folder lazily ensured on first listing.
     if await permission_service.is_workspace_member(workspace_id, current_user["id"]):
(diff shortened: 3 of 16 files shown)


henry-dowling:
### Follow-up: pinned folder now honored on the event-stream path (closes the Codex gap)

The first version honored a repo's pinned folder only for agents with a session-start hook (Claude). Agents without one — Codex, cursor, gemini, opencode, openclaw — create the session row from their **first streamed event**, so the pin was ignored and the session still landed in **Default**.

This commit streams `session_folder_id` on every event through the whole chain — plugin `push_event` → `POST /sessions/events` → `memory_service.push_event` → `upsert_session` — so the pin reaches whichever call first creates the row, for **every** agent. So `stash sessions use-folder` now works regardless of agent.

To make streaming the pin on every turn safe, `upsert_session` no longer touches `session_folder_id` on conflict (it was `COALESCE`, which re-homed a `NULL`). The folder is now genuinely **set once at insert**, so a user's explicit move-to-root can't be undone by the next streamed event.

**New tests:**
- Pinned event-stream session lands in the pinned folder (not Default).
- A streamed pin does **not** re-home a session after an explicit move-to-root.

PR #677, opened 2026-06-23
Unified live search bar + searchable filters; page-search permissions in SQL

ColemanDuPlessie:
## Changes

Redid search UI and UX. A brief summary of changes:
- Unified search into the global top-of-page search bar (previously, there were two search bars on the search page)
- Removed "Workspace" and "Skill" filter dropdowns as obsolete
- Greatly reduced size of remaining dropdowns and moved them to above search results, rather than beside them
- Added a highlight in each search result showing where the searched text appears
- Removed an inefficiency in search where we were fetching more results than necessary and then checking if we have read permissions: now, we use permission_service.readable_content_condition to avoid having to loop over results and await the result of permission_service.check_access for each one.

AI-generated notes preserved below this line
------------------------------------------------------

### Frontend — unified, live search
- **One search bar.** On `/search`, the app-shell top bar becomes a live text input (`InlineSearchInput`) instead of the button that opened the command palette. Typing writes a debounced `q` into the URL; the search page reads `q` and re-runs the search in real time. The command palette stays reachable via the `⌘K` affordance and the global shortcut.
- **Filters moved into a top bar.** The old left sidebar of stacked dropdowns (Workspace / Skill / Folder / Page / Content) is replaced by a single horizontal row of pill-style filters above the results.
- **Type-to-filter dropdowns.** `CustomSelect` gains a `searchable` mode: an in-menu search box filters options, with keyboard navigation operating over the filtered view. Used for the Folder/Page filters so long lists are usable.
- **Result `detail` is now `ReactNode`**, allowing richer (e.g. highlighted) result snippets.

### Backend — page search permissions in SQL
- Drops `_filter_readable`, which fetched `limit * 3` rows and then called `permission_service.check_access` per row (an N+1). Both `search_pages_fts` and `search_pages_vector` now push the readability check into the query via `permission_service.readable_content_condition`, so the DB returns only readable rows and we fetch exactly `limit`.

## Testing
- `frontend`: added `CustomSelect.test.tsx` covering the new searchable behavior — 5/5 passing (`npx vitest run src/components/CustomSelect.test.tsx`).

## Screenshots

**Live search with highlighted matches** (searching "apple") — unified top bar, pill filters, results ranked by relevance:

Search results for apple

**Empty state** — search bar with ⌘K affordance and the filter pill row above results:

Empty search page


backend/services/files_tree_service.py
```diff
@@ -1172,22 +1156,22 @@ async def search_pages_vector(
...
         "SELECT id, owner_user_id, folder_id, name, content_markdown, content_html, "
         "content_type, html_layout, metadata, "
         "created_by, updated_by, created_at, updated_at, "
         "1 - (embedding <=> $2) AS similarity "
-        f"FROM pages WHERE owner_user_id = $1 AND {_PAGE_FILTER} "
-        "AND embedding IS NOT NULL "
-        "ORDER BY embedding <=> $2 LIMIT $3",
-        owner_user_id,
-        query_embedding,
-        limit * 3,
```
henry-dowling, line 1185: This thing was incredibly jank, so glad to get rid of that!

frontend/src/app/search/page.tsx
```diff
@@ -159,9 +159,14 @@ function SearchPageInner() {
 
   const sourceName = user?.display_name ?? "You";
 
-  const handleSearch = useCallback(async () => {
-    const q = query.trim();
-    if (!q) return;
+  const handleSearch = useCallback(async (rawQuery?: string) => {
+    const q = (rawQuery ?? query).trim();
```
henry-dowling: What's the difference between `rawQuery and query` exactly? I believe the `?? query` branch will never execute.

frontend/src/components/AppShell.tsx
```diff
@@ -221,6 +221,73 @@ function TopSearchButton({
   );
 }
 
+// On the /search route the top bar is a live text input that drives the
+// results in real time (it writes the query into the URL, which the search
+// page reads). The command palette stays reachable via the ⌘K affordance and
```
henry-dowling: Pressing cmd-k won't open up the search bar on /search, right? It'll open up the command palette. That's confusing IMO. 

My recommendation is to remove the affordance so if you're on the search page, you use the search bar. We'll make this whole system less jank later on in the addendum to your internship project that I'm going to share in slack tonight

henry-dowling (changes requested):
Nice work! Left a couple requests for changes but overall this is good work. Please take a look at the comments I left and then once we've addressed them we can merge

frontend/src/app/search/page.tsx
ColemanDuPlessie: Good catch, I looked and it's no longer ever actually used, so I deleted it (and some code that sets query elsewhere in the file).

frontend/src/components/AppShell.tsx
ColemanDuPlessie: Makes sense. I just updated it so that cmd-K now just focuses the search bar on the search page

PR #683, opened 2026-06-25
feat(vfs): cap command output so a single read can't flood an agent's context

henry-dowling:
## Why

The `stash vfs` shell is an agent-facing surface, and agents read command output into a finite context window. But every command returned **unbounded** output — `cat` of a long session transcript, or `ls` of a 1000-message Slack channel, dumped the entire thing into the agent's context. (Inspired by how Mesa's `path_content` pages large reads — same problem, agent-context protection.)

## What

One output budget applied at the shell boundary, covering **every** read command (`cat`, `ls`, `grep`, `tree`, `find`) in a single chokepoint inside `run()`:

- `MAX_OUTPUT_LINES = 500` caps the **final** result returned to the agent.
- A loud, copy-pasteable footer tells the agent how to page the rest:
  ```
  [showing lines 1-500 of 1000 — page with: ls /sources/slack/general | sed -n '501,1000p']
  ```
- The cap is **terminal-only**: pipeline stages still produce full output, so `cat big | sed -n '501,1000p'` returns the next page intact instead of a doubly-truncated fragment. Paging is fully stateless — each page is a self-contained call.

No new flags, no hidden state — it leans on the `sed -n` paging the shell already ships.

## Design notes

- Chose a single global output budget over per-command `--limit/--offset` flags: one codepath, protects `grep`/`tree`/`find` too, and one paging idiom (`sed`) instead of two.
- Cap is **line-based** so paging maps cleanly to `sed -n 'N,Mp'`. The edge case it doesn't cover is a single pathologically long line (e.g. a one-line giant JSON blob); a secondary byte ceiling would be the follow-up, left out to keep one codepath.
- The footer uses the **no-space** `'N,Mp'` form, because this shell's `_sed` only accepts that (not GNU sed's `'N,M p'`). A test enforces the hint is runnable in this shell.

## Tests

Three new tests in `cli/tests/test_app_vfs.py`:
- `_cap_output` truncates long output with a correct footer (range, total count, runnable command).
- Output within budget passes through untouched.
- The cap is terminal-only: `cat big | sed -n '3,4p'` returns the next page intact, proving pipelines aren't capped and the footer's advice actually works.

All `cli/tests/test_app_vfs.py` tests pass except one **pre-existing** failure (`test_app_vfs_surfaces_backend_timestamps_and_marks_unknown`) that references the `/me/files` path removed in c8e391ea — confirmed failing independently of this change.

### DIFF
--- a/cli/app_vfs.py
+++ b/cli/app_vfs.py
@@ -13,6 +13,11 @@ from datetime import datetime
 from .client import StashError
 from .mount import MountError, StashVfsModel
 
+# Agents read command output into a finite context window. Cap what a single
+# command returns so `cat` of a long transcript or `ls` of a 1000-message Slack
+# channel can't flood it; the footer tells the agent how to page the rest.
+MAX_OUTPUT_LINES = 500
+
 
 @dataclass
 class VfsCommandResult:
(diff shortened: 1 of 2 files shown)


henry-dowling:
nvm this is dumb- cc harness already truncates

PR #704, opened 2026-07-04
Surface dead federated sources instead of silently returning no results

henry-dowling:
make two fixes that cause sources to silently fail when you try to search them

## Problem

A second connected Gmail mailbox (`henry@ferganalabs.com`) was fully populated in the VFS but returned **zero** results in search, while the other mailbox worked. Root cause was silent-fallback bugs, not a crawl failure:

1. **Search swallowed the auth error.** Gmail search is federated live to the provider. When the ferganalabs OAuth token expired (`expires_at` ~17 days in the past, no longer refreshing), `get_valid_token` raised 401 → `_federated_search`'s unscoped fan-out caught it, logged a warning, and returned `[]`. A dead connection was indistinguishable from "no matches."
2. **Status reported `connected: true`.** `/integrations/gmail/status` derived `connected` from the mere existence of a token row, so a dead account looked healthy in the UI.

Both are the "silent fallback" pattern `CLAUDE.md` forbids — fail loud instead.

## Changes

- **`_federated_search`** (`backend/services/source_service.py`): in the unscoped fan-out, a provider error now returns a typed marker `{source, source_name, error, needs_reconnect}` instead of `[]`. The rest of the search stays alive; the dead source is surfaced. Only the classified (HTTP-mapped) message is exposed — an unmapped exception gets a generic label so raw exception text (token/query) can't leak. Scoped search is unchanged (still raises an explicit HTTP error).
- **`status` / `list_connections`** (`backend/integrations/storage.py`): each account is now actively verified by attempting to obtain a valid token, adding a per-account `needs_reconnect` boolean.
- **`IntegrationAccountItem`** (`backend/integrations/router.py`): declares `needs_reconnect` so the flag survives serialization on `GET /api/v1/integrations` (the payload the UI reads) — without this the banner never rendered.
- **Frontend** (`integrations/[provider]/page.tsx`, `lib/integrations.ts`): `IntegrationAccount.needs_reconnect` type + a reconnect banner on the per-integration page when an account is connected-but-dead.

## Tests

- `test_gmail_status_flags_dead_account_for_reconnect` — dead account reports `needs_reconnect: true`; healthy one `false`.
- `test_integrations_list_exposes_needs_reconnect` — the flag survives the router response model.
- `test_unscoped_search_surfaces_dead_federated_source` — a 401 from a federated source surfaces a `needs_reconnect` marker instead of vanishing.
- `test_federated_search_logs_only_failure_metadata` — updated for the new marker; asserts it does **not** leak the token/query from an unmapped exception.

Backend suites passing (`test_sources.py`, `test_integration_sources.py`, `test_gmail_multi_account.py`, callback/token-refresh); `ruff` clean; frontend `tsc`/`eslint` clean on changed files. Product screenshot of the reconnect banner + end-to-end verification in the thread below.

### DIFF
--- a/backend/integrations/router.py
+++ b/backend/integrations/router.py
@@ -95,6 +95,7 @@ class IntegrationAccountItem(BaseModel):
     scopes: list[str]
     expires_at: str | None = None
     connected_at: str | None = None
+    needs_reconnect: bool = False
 
 
 class ProviderListItem(BaseModel):
--- a/backend/integrations/storage.py
+++ b/backend/integrations/storage.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 from datetime import UTC, datetime, timedelta
 from uuid import UUID
 
+import httpx
 from fastapi import HTTPException
 
 from ..database import get_pool
@@ -199,6 +200,18 @@ async def revoke_stored(
             pass
 
 
+async def _account_needs_reconnect(user_id: UUID, provider: str, account_key: str) -> bool:
+    """Actively verify the stored token still works, refreshing if needed. A
+    connection row survives long after its OAuth grant dies (revoked access, an
+    expired refresh token), so only attempting to obtain a valid token tells a
+    live account apart from a silently dead one."""
+    try:
+        await get_valid_token(user_id, provider, account_key)
+        return False
+    except (HTTPException, httpx.HTTPError):
+        return True
+
+
 async def status(user_id: UUID, provider: str) -> dict:
     pool = get_pool()
     rows = await pool.fetch(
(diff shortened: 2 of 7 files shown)


henry-dowling:
### Product screenshot — reconnect banner (Fix #2)

Captured against the local stack with a seeded dead Gmail account (`henry@ferganalabs.com`: expired token, no refresh token) alongside a healthy one (`htdowling@gmail.com`). The per-integration page now shows a reconnect banner ("`henry@ferganalabs.com is connected but its access expired — search returns nothing until you reconnect.`" + a **Reconnect** button) instead of a bare "2 accounts connected".

_(Screenshot to be dropped into this thread — binary image upload isn't available through the `gh` CLI.)_

**End-to-end verification performed:**
- `GET /api/v1/integrations` returns `needs_reconnect: true` for the dead account, `false` for the healthy one (confirmed against the running backend).
- The banner renders only for the stale account; the healthy mailbox is unaffected.
- Backend suites on a fresh DB: `test_gmail_multi_account.py` (4), `test_sources.py` + `test_integration_sources.py` + callback/token-refresh (122) — all passing. `ruff` clean; frontend `tsc`/`eslint` clean on changed files.

PR #708, opened 2026-07-05
Disclose federated search truncation instead of silently capping

henry-dowling:
warn that we're truncating results in gmail search

# Slop
## Problem

Gmail search is federated live and caps at `SEARCH_LIMIT = 25` results per mailbox. We returned that top slice as a bare list, so a caller — the **agent's `_search` tool** especially — couldn't tell "these are all your matches" from "here are the first 25 of hundreds." It would confidently answer as if it had seen everything.

And we were throwing away the exact signal to fix it: Gmail's `messages.list` returns `nextPageToken` (⟺ more matches exist) and `resultSizeEstimate` (rough total), both discarded in `_list_message_refs`.

## Changes

- **Gmail indexer**: `_list_message_refs` now returns `(refs, nextPageToken, resultSizeEstimate)`; `search_gmail` returns `{hits, truncated, estimated_total}` where `truncated = bool(nextPageToken)` — the authoritative "there is more" signal, **not** a `len(hits) == limit` guess (which can't work here, since the cap is *below* the requested limit).
- **`_federated_search`**: normalizes Gmail's richer return and appends a trailing truncation marker `{source, source_name, truncated, returned, estimated_total}` when the provider capped. Consistent with the error-marker pattern from #704.
- **CLI `stash search`**: renders the truncation marker as a `… showing first N of ~M matches` note, and the (previously unhandled) error marker as a warning — instead of blank rows.
- **Frontend** per-integration search box: filters marker entries out of the hit list and shows a `Showing the first N of ~M matches` notice.

## Scope

Gmail-only for now (it has a clean `nextPageToken`). The other federated providers (Jira/Asana/Drive/Linear) share the same silent-cap and still return plain lists — a fast-follow can add their per-provider "more exists" signal through the same marker path. Raising `SEARCH_LIMIT` / paginating is a separate lever, orthogonal to *disclosing* the cut.

## Tests

- `test_search_gmail_reports_truncation_from_next_page_token` — `nextPageToken` → `truncated: true`, `None` → `false`; surfaces `resultSizeEstimate`.
- `test_search_appends_truncation_marker_when_provider_caps` / `..._omits_..._when_not_capped` — marker wiring through `search_all`.
- `test_search_renders_error_and_truncation_markers` (CLI) — both marker kinds render.
- Existing gmail index test updated for the new `_list_message_refs` tuple.

Full lint CI-equivalent (`ruff check` + `ruff format --check` on `backend/ cli/`) clean; frontend `tsc`/`eslint` clean on changed files. Backend suites green (one **pre-existing, unrelated** Gong failure — `test_gong_index_success...`, fails on `origin/main` too).

**Real-world evidence this fires:** live prod queries against the reconnected `henry@ferganalabs.com` mailbox already return exactly 25 for `Docusign`/`Coleman`/`workspace-noreply` (i.e. capped) — those will now surface the truncation notice on deploy. UI screenshot in the thread.

### DIFF
--- a/backend/integrations/gmail/indexer.py
+++ b/backend/integrations/gmail/indexer.py
@@ -75,13 +75,21 @@ async def _list_message_refs(
     client: httpx.AsyncClient,
     query: str,
     limit: int,
-) -> list[dict]:
+) -> tuple[list[dict], str | None, int | None]:
+    """Returns (message refs, nextPageToken, resultSizeEstimate). A nextPageToken
+    means Gmail matched more than we fetched — the authoritative "there is more"
+    signal; resultSizeEstimate is Gmail's rough total-match count."""
     resp = await client.get(
         MESSAGES_URL,
         params={"q": query, "maxResults": min(limit, 100)},
     )
     resp.raise_for_status()
-    return resp.json().get("messages", []) or []
+    body = resp.json()
+    return (
+        body.get("messages", []) or [],
+        body.get("nextPageToken"),
+        body.get("resultSizeEstimate"),
+    )
 
 
 async def _get_message_metadata(client: httpx.AsyncClient, message_id: str) -> dict:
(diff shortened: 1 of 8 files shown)


henry-dowling:
### Product screenshot — truncation notice (frontend)

Per-integration search box showing 25 results capped with the disclosure line at the bottom: **"Showing the first 25 of ~213 matches — narrow your search to see more."**

_(Rendered against the local stack. Live Gmail OAuth isn't configured locally, so the scoped `/sources/search` response was stubbed with the exact backend marker shape — `{truncated, returned, estimated_total}` — to drive the real component. Screenshot to be dropped into this thread; `gh` can't upload binary attachments.)_

**CLI equivalent** (`stash search`), covered by `test_search_renders_error_and_truncation_markers`:
```
… Gmail (henry@ferganalabs.com): showing first 25 of ~213 matches — narrow the query to see more.
```
The agent's `_search` tool gets the same marker in its JSON tool-result, so it can no longer present the top 25 as the full set.

PR #725, opened 2026-07-07
Support scanned PDFs via Claude vision OCR

henry-dowling:
## Problem

A scanned PDF has no embedded text layer, so pypdf extraction yields nothing — the file row ends up with `extracted_text = NULL`, never gets embedded, and is invisible to the agent's knowledge base (`read_file` returns empty text, semantic search never surfaces it).

## What this does

When `extract_text` finds no text in a PDF, `extract_one` now treats it as a scan and transcribes it with Claude vision via the already-configured `ANTHROPIC_FAST_MODEL` tier (Haiku by default) — no new system dependencies, no Docker changes; the `anthropic` SDK was already a backend dep.

- **`backend/services/pdf_ocr.py`** (new): sends the PDF to the API as a base64 `document` block. Pages go up in parallel 10-page chunks (pypdf slicing), capped at `MAX_OCR_PAGES = 100` with an explicit `[OCR stopped at page 100 of N]` marker appended to the stored text — bounded spend, never a silent cap.
- **Fail loud**: missing API key or API errors raise into the existing redacted-error path (`Extraction failed: <ExceptionType>`), so the row retries up to 3 attempts and then reads `failed` — never a `done` row with silently-empty text.
- **`extract_one`**: OCR fires only when embedded extraction found nothing *and* the file is a PDF; PDFs with a text layer never hit the API.
- **Child timeout** raised 180s → 600s: OCR calls run with a 120s per-request timeout plus SDK retries.

Once transcribed, the text flows through the existing pipeline unchanged: `embed_stale` flips on, the reconciler embeds it, and the agent's `read_file`/search see it.

Cost: roughly $0.005–0.01 per scanned page on the fast tier.

## Testing

- 7 new tests in `backend/tests/test_pdf_ocr.py`: chunk sizing, page-cap marker, textless-PDF→OCR wiring, text-layer PDFs and non-PDFs never trigger OCR, and OCR failures land in `extraction_error` for retry.
- Full backend suite: 697 passed; the 4 failures in `test_agent_oauth.py` also fail on a clean tree (pre-existing, unrelated).
- `ruff check` / `ruff format --check` clean.
- ✅ Live-API smoke test passed against the real API with the production key (see PR comment): an image-only PDF routes to OCR and every figure round-trips exactly.

## Known gap (out of scope)

The Google Drive indexer (`backend/integrations/google/indexer.py`) uses the same `extract_text` but does not get OCR — scanned PDFs synced from Drive still won't be transcribed.

### DIFF
--- a/backend/services/file_extraction.py
+++ b/backend/services/file_extraction.py
@@ -39,6 +39,11 @@ _OOXML_DRAWING_NS = "{http://schemas.openxmlformats.org/drawingml/2006/main}"
 _OOXML_WORD_NS = "{http://schemas.openxmlformats.org/wordprocessingml/2006/main}"
 
 
+def is_pdf(content_type: str) -> bool:
+    ct = (content_type or "").lower()
+    return ct == "application/pdf" or ct.endswith("/pdf")
+
+
 def _extract_pdf_embedded(content: bytes) -> str:
     if not _HAS_PYPDF:
         return ""
@@ -155,7 +160,7 @@ def extract_text(content: bytes, content_type: str) -> str | None:
     try:
         ct = (content_type or "").lower()
 
-        if ct == "application/pdf" or ct.endswith("/pdf"):
+        if is_pdf(ct):
             text = _extract_pdf_embedded(content)
             return _sanitize_for_postgres(text) if text else None
 
--- a/backend/services/pdf_ocr.py
+++ b/backend/services/pdf_ocr.py
@@ -1,0 +1,97 @@
+"""OCR for scanned PDFs via Claude vision.
+
+A scanned PDF has no embedded text layer, so pypdf yields nothing.
+This module sends the PDF itself to the configured fast-tier model
+(`ANTHROPIC_FAST_MODEL`), which reads the page images and transcribes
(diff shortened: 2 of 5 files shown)


henry-dowling:
✅ Live-API smoke test now complete (was blocked on a valid key; used the production key from Render).

Generated an image-only PDF (no text layer), confirmed `extract_text` returns `None` (routes to the OCR path), then ran `ocr_pdf` against the real API with `claude-haiku-4-5`:

```
embedded extraction result: None  (None means scanned -> OCR path fires)
--- OCR output ---
Heavi Industries Q3 Report
Revenue grew 42 percent year over year.
The flux capacitor division shipped 1,847 units.
Page two of the scan.
Operating margin reached 31 percent.
We expect continued growth through fiscal 2027.
------------------
PASS: all expected content transcribed
```

All figures (42%, 1,847 units, 31%, 2027) round-tripped exactly. Note: production has `ANTHROPIC_API_KEY` set but no `ANTHROPIC_FAST_MODEL` override, so OCR will run on the default `claude-haiku-4-5` — no Render config changes needed to ship this.

PR #736, opened 2026-07-07
Fix tab scrolling: bounded flex column for tab bodies

samzliu:
## Problem

Content in workbench tabs frequently could not scroll (reported on the Memory-curator settings tab, but general). Two root causes, one shared origin:

1. The per-tab wrapper in `workbench.tsx` was a plain `h-full` **block** inside the tab host's `overflow-hidden`. Every tab body whose root relies on `flex-1 overflow-y-auto` — page, session, folder, skill, file — had `flex-1` silently ignored (no flex parent), grew to content height, and was clipped with no scrollbar.
2. The `agent-config` (curator settings) and `tool` tabs had no height or scroll container at all.

The codebase already documented the problem in a comment ("SessionsPage's root is flex-1 and needs a bounded flex-column parent to scroll … the tab host isn't") and hand-rolled the fix twice (sessions-home wrapper, AgentChatTab).

## Fix

- Per-tab wrapper → `flex h-full min-h-0 flex-col` (one line, fixes the whole flex-1 family; the sessions-home special case is now redundant and removed). No scroller added at the host level — each body still owns its own scroll, per the existing comment about contenteditable paint.
- `agent-config` and `tool` tabs get a `min-h-0 flex-1 overflow-y-auto` scroller, mirroring what AgentChatTab already does for the identical `AgentConfigPanel`.

## Verification

- Locally seeded a 40-item folder: before, the folder tab clipped with no scrollbar; after, it scrolls (screenshot in comment — view scrolled to bottom, `scrollTop` moves).
- Page tab, sessions-home render clean, no console errors.
- Frontend: 186 tests pass, eslint clean, build passes.

### DIFF
--- a/frontend/src/components/workspace/tab-body.tsx
+++ b/frontend/src/components/workspace/tab-body.tsx
@@ -92,25 +92,28 @@ export default function TabBody({ tab }: { tab: WorkbenchTab }) {
   if (tab.kind === "page") return <PageClient pageId={tab.refId} />;
   if (tab.kind === "file") return <FileClient fileId={tab.refId} />;
   if (tab.kind === "table") return <TableClient tableId={tab.refId} embedded />;
-  if (tab.kind === "sessions-home")
-    // SessionsPage's root is `flex-1` and needs a bounded flex-column parent to
-    // scroll (the standalone route's <main> is one; the tab host isn't).
-    return (
-      <div className="flex h-full min-h-0 flex-col">
-        <SessionsPage />
-      </div>
-    );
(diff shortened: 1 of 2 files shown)


samzliu:
Screenshot — 40-item folder tab scrolled to the bottom (previously clipped at the fold with no scrollbar): https://app.joinstash.ai/f/183f8d7b-bd59-40e0-a237-eaddb9ece403

henry-dowling (approved):
lgtm!

PR #766, opened 2026-07-10
CLAUDE.md: worktree isolation rules for concurrent agent sessions

henry-dowling:
## Why

On 2026-07-10, five concurrent Claude sessions in separate worktrees collided on every piece of shared machine state (reconstructed from local transcript forensics):

- **Ports**: repeated EADDRINUSE on 3456–3462 as sessions improvised "nearby" ports and raced each other (3459 alone was grabbed at by three sessions).
- **Database**: one session ran `docker compose down`, tearing down the shared Postgres/Redis under every other worktree.
- **Browser**: three sessions drove the shared `agent-browser` daemon without `--session`; one session's `open localhost:3461/...` stole another session's tab mid-QA-flow.

## What

A new `### Worktree isolation` section in CLAUDE.md (AGENTS.md symlinks to it, so Codex/OpenCode pick it up too):

1. `export AGENT_BROWSER_SESSION="$(basename "$PWD")"` before any `agent-browser` use — the default session is one shared Chrome.
2. Primary checkout owns default ports; secondary worktrees pin a pair once in `.env` (`BACKEND_PORT`/`FRONTEND_PORT`) and reuse it — no probing for free ports at session time.
3. Never `docker compose down`/`restart` the shared stack from a secondary worktree; run a per-worktree `stash-dev-pg-<worktree>` pgvector container instead.
4. Kills by PID or exact port only (folds the existing kill-scoping rule into the section; the self-sufficiency bullet now references it).

Docs-only change.


henry-dowling:
Moving these to user-level ~/.claude/CLAUDE.md instead — the multi-worktree workflow is machine-specific, not a team convention.

PR #779, opened 2026-07-11
VFS endpoint: cap concurrent scripts and add a wall-clock budget

henry-dowling:
## Why

**Incident, 2026-07-09 20:57 UTC:** a customer's production agent fired ~20 concurrent `POST /api/v1/me/vfs` scripts. Each request builds a full `StashVfsModel` (every listing walked) and holds every document its command reads in memory. Usage raced from the 272MB baseline toward the 512MB instance limit; the kernel killed the API process (Render: "Instance restarted"), and since one instance serves the whole product, **every customer got 502s for ~8 seconds and all in-flight requests were dropped**. The 502 bodies were Render's error page answering for the dead upstream.

The endpoint had a per-request read-count budget (`MAX_DOCUMENT_READS`) but nothing bounded concurrency or wall-clock time — and long scans (we logged an 81s request) are exactly what stack requests into a pile-up.

## What

Two guards in `vfs_service`, both following the existing budget pattern:

- **`MAX_CONCURRENT_SCRIPTS = 4`** — past the cap, the request fails immediately with **429 + `Retry-After`**. Deliberately no server-side queue: a queue would hold the memory the cap exists to not hold. Agent callers retry naturally.
- **`MAX_SCRIPT_SECONDS = 60`** — a monotonic deadline checked on every nested read, raising the existing `VfsBudgetExceeded` (**413**). A runaway scan dies at its next read past the deadline instead of holding a slot (and its memory) indefinitely.

The counter is checked and bumped on the event loop with no await in between, so no lock; the `finally` guarantees slot release on success, budget abort, or crash.

After this, the failure mode for a burst is "one agent sees 429/413" instead of "the API dies for everyone." Instance count ≥2 and per-key rate limiting are follow-ups from the same incident review.

## Testing

- New tests in `backend/tests/test_vfs.py`: over-cap request → immediate 429 with `Retry-After`; slot released after a run (cap=1, two sequential requests both succeed); past-deadline command aborts with 413 naming the time limit.
- `pytest backend/tests/test_vfs.py` — 12 passed (9 pre-existing + 3 new).

### DIFF
--- a/backend/routers/vfs.py
+++ b/backend/routers/vfs.py
@@ -11,7 +11,7 @@ from stashvfs import MountError
 
 from ..auth import get_current_user
 from ..services import vfs_service
-from ..services.vfs_service import VfsBudgetExceeded
+from ..services.vfs_service import VfsBudgetExceeded, VfsBusy
 
 router = APIRouter(prefix="/api/v1/me/vfs", tags=["vfs"])
 
@@ -47,3 +47,5 @@ async def run_vfs(
         raise HTTPException(status_code=400, detail=str(e)) from e
     except VfsBudgetExceeded as e:
         raise HTTPException(status_code=413, detail=str(e)) from e
+    except VfsBusy as e:
+        raise HTTPException(status_code=429, detail=str(e), headers={"Retry-After": "2"}) from e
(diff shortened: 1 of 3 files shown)


henry-dowling:
Superseded by #997, which reapplies the overload protection on current main, preserves the newer VFS behavior, and adds true concurrency plus failure-release coverage.

PR #789, opened 2026-07-13
Curator reads documents via reader subagents, RLM-style

henry-dowling:
## Problem

The Memory curator runs as a single headless agent session, but a customer corpus can exceed its context window. Heavi's bootstrap run (2026-07-10, 40+ artifacts, ~1.4MB extracted text vs a ~200K-token window) showed exactly what happens then — the event log for that run shows the curator read:

| Transcript | Actual read |
|---|---|
| Brandon Walz #2 (69KB) | `head -20` |
| Brandon Walz #3, Dan Larlee #2 | `head -10` each |

…then wrote complete-looking annotation pages whose "key facts" were literally *"File size (69KB) suggests substantial domain discussion. (inferred)"*, marked its read-the-sources todo completed, and the next run's changelog declared all artifacts annotated. The hollow pages are permanently laundered as done.

This is the "Algorithm 2" anti-scaffold from [Recursive Language Models](https://arxiv.org/abs/2512.24601) (Zhang, Kraska, Khattab): corpus content pulled into the root's history until window pressure forces lossy heuristics.

## Change

Restructure `render_curator_prompt` around the RLM discipline. No plumbing changes — scheduled curator runs already execute headless Claude Code with no tool restrictions, so the Task tool is available; the prompt just never imposed the discipline.

- **Peek, don't accumulate:** the root may print bounded document snippets to triage (category choice, duplicate spotting), but must not fill its window with document bodies — full ingestion happens in subagents.
- **One reader subagent per document:** reads the ENTIRE document in its own disposable context (chunked `sed -n` for large files), writes/updates the wiki page itself, and returns only `page_id | one-line summary | topic tags`. A small document the root fully read during triage is the exception — it may write that page directly.
- **No facts from partial reads:** a document that can't be fully read this run becomes an explicit stub page plus an `INCOMPLETE` entry in `Log` — next-run work, never a confident shell.
- **Root's job = inventory → dispatch → weave:** categories, cross-links, index, and `Log` from digests. Chat history still curated inline (it arrives via `stash changes` and is small).
- Worked bootstrap example in the prompt, since the paper finds in-context decomposition examples strongly improve the first decomposition attempt.

Root context now scales with the number of documents (~300 chars each), not the corpus.

## Testing

- `pytest backend/tests/test_curator.py` — 20 passed, including new `test_curator_prompt_reads_documents_via_subagents` pinning the accumulate ban, whole-document rule, and INCOMPLETE protocol.
- `ruff check` / `ruff format --check` clean.
- Existing accounts pick the new prompt up on their next scheduled run; suggest validating with a fresh bootstrap on the heavi-test account and diffing transcript-page depth against prod's hollow pages before any backfill on Heavi's real account.

### DIFF
--- a/backend/services/prompts.py
+++ b/backend/services/prompts.py
@@ -112,7 +112,15 @@ def render_curator_prompt(memory_folder_id: str, since: str | None) -> str:
     Structured on Karpathy's LLM-wiki pattern: raw sources (the user's stash
     activity) are immutable inputs, the wiki under the Memory folder is the
     compiled, compounding artifact, and this prompt is the schema — page
-    types, linking rules, and the ingest + lint workflows."""
+    types, linking rules, and the ingest + lint workflows.
+
+    Reading is recursive, RLM-style (arXiv:2512.24601): the corpus can exceed
+    one context window, so the root agent handles sources symbolically (paths
+    and metadata only) and each artifact is fully read inside a disposable
+    reader subagent that writes the page and returns a constant-size digest.
+    A bootstrap run once skimmed transcripts with `head -20` and published
+    confident-looking pages whose "facts" were inferred from byte sizes —
+    this structure exists so that can't recur."""
     window = (
         f"the changes since {since}"
         if since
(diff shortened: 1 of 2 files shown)


samzliu:
Soem comments: 
- How do we deal with 1000 page PDFs? Is the right behavior that the agent has to read everything or would reading only the first 50 pages of this be enough to "index" it in the wiki?
- We should probably tell the agent that the sub-agents can be cheaper models to save on compute costs?
- Part of the problem in addition to it being lazy seems that the facts it draws out are not useful (e.g. the size of the file doesn't matter). There's probably some prompting there we can do to make it actually extract useful information

PR #795, opened 2026-07-13
Clip API: save webpages and PDFs from the browser extension

samzliu:
First slice of the commonplace-book feature (plan: clipper → URL imports → bookmark/tabs imports → X bookmarks → Instagram saves → extension).

## What
- `POST /api/v1/me/clips/page` `{url, html, title?}` — the extension sends the captured, logged-in DOM; one server-side extractor (**trafilatura**) turns it into a markdown page in a **Clips** root folder, with `source_url`/`clipped_at` in page metadata plus a `> Clipped from <url>` header in the body so FTS and the curator see the URL.
- `POST /api/v1/me/clips/file` — binary clips (PDFs) ride the existing `ingest_bytes` path into Clips; new `files.source_url` column (migration 0144) records provenance.
- Extraction failure = loud 422; no fallback extractor. Re-clips get " (2)" names — user edits are never clobbered.

## Why
Clips are uploads, not a new entity: the extension is just another upload client, so existing FTS, embeddings, and the nightly curator pick clips up with zero extra machinery.

## Tests
`backend/tests/test_clips.py` — extractor unit tests, Clips-folder landing + metadata, 422 fail-loud, duplicate-name suffix, file source_url. Full suite green locally (pre-existing local-env failures fixed separately: Postgres timezone + S3 env).

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -25,6 +25,7 @@ from .routers import (
     batch,
     billing,
     bulk_export,
+    clips,
     collab,
     demo,
     discover,
@@ -124,6 +125,7 @@ app.include_router(tables.me_router)
 app.include_router(tables.router)
 app.include_router(files.me_router)
 app.include_router(files.canonical_router)
+app.include_router(clips.router)
 app.include_router(batch.router)
 app.include_router(transcripts.router)
 app.include_router(aggregate.router)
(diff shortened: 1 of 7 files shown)


samzliu:

PR #796, opened 2026-07-13
URL-import worker: async clips for YouTube, arXiv, PDFs, and links

samzliu:
Stacked on #795. Second slice of the commonplace-book feature.

## What
- `url_imports` + `import_batches` tables (migration 0145): job state for URL-only clips, mirroring the `files.extraction_status` lifecycle (claim guard, attempts, Beat sweep for stragglers).
- One server-side URL router (`clip_router.py`): YouTube → transcript page via **yt-dlp** (manual captions beat auto, English beats original language — no cascade, loud `TranscriptUnavailable`); arXiv `/abs/` → the paper PDF; any PDF → file clip; HTML → article page; anything else fails loud per row.
- `POST /clips/page` now returns **202 + import_id** for URLs whose content isn't in the posted DOM (YouTube/arXiv); `GET /clips/{import_id}` reports status/error/result.
- Batched worker tasks (100 URLs per Celery task, concurrency 8 inside) so a big import can't starve other queue work — sized for the 10k-bookmark imports in the next PR.

## Verified
- 13 new tests in `test_url_imports.py` (classification, caption parsing, worker flows, fail-loud rows, claim idempotency).
- Real-world smoke test: `fetch_transcript("…jNQXAC9IVRw")` returns the actual "Me at the zoo" transcript.

### DIFF
--- a/backend/celery_app.py
+++ b/backend/celery_app.py
@@ -24,6 +24,7 @@ celery = Celery(
     backend=settings.REDIS_URL,
     include=[
         "backend.tasks.extraction",
+        "backend.tasks.clips",
         "backend.tasks.drive_extraction",
         "backend.tasks.embeddings",
         "backend.tasks.linear_tickets",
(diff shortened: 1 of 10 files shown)


samzliu:

PR #797, opened 2026-07-13
Bookmark and tab imports: mirror the browser tree, built for 10k+

samzliu:
Stacked on #796. Third slice of the commonplace-book feature.

## What
- `POST /api/v1/me/imports/bookmarks` — upload a Netscape-format bookmarks.html; the browser's folder tree is mirrored under `Clips/Bookmarks/…` and each URL becomes a `url_imports` row fetched out-of-band (article extraction / PDF / YouTube routing from #796). Cap 25,000 URLs, loud 413 above.
- `POST /api/v1/me/imports/tabs` — "clip all open tabs": up to 200 URLs into `Clips/Tabs/<date>/`.
- `GET /api/v1/me/imports/{batch}` — `{total, done, failed, pending, failures[:50]}` polling endpoint.
- Parser is stdlib `html.parser`, tolerant of the unclosed `</DT>`/`</A>` tags real exports produce; drops `javascript:`/`place:` schemes; dedupes.

## Scale (10k target, not edge case)
Synthetic 10k-bookmark import through the real endpoint + worker code: **accepted in 0.2s as exactly 100 batches; 10,000/10,000 processed; progress endpoint accurate; 10,000 pages created.** Batched tasks (100 URLs/task, concurrency 8) keep the queue shallow so periodic work interleaves.

## Tests
`backend/tests/test_bookmark_imports.py` — parser (nested folders, ragged HTML), tree mirroring assertions to the Clips root, empty/oversized rejections, tabs dedupe + dated folder, per-URL failure reporting, owner scoping.


samzliu:

PR #798, opened 2026-07-13
X bookmarks as a synced, archived source

samzliu:
Stacked on #797 (for migration numbering only — no code dependency). Fourth slice of the commonplace-book feature.

## What
- New `twitter_bookmarks` source type on the existing Twitter OAuth provider (`bookmark.read` scope + `BOOKMARKS_URL` already existed): `index_twitter_bookmarks` pulls one page of 100 per 6-hour sync (X free tier allows 1 bookmarks call/15 min) and upserts rendered posts into a new `twitter_bookmark_docs` content table (migration 0146, `drive_documents` shape).
- **Archive semantics**: upsert-only, no `remove_missing_documents` — a bookmark stays in Stash after it's un-bookmarked or ages past the API's first page.
- Content table is in `CONTENT_TABLES`, so FTS via `search_documents` and embedding reconciliation come for free; docs appear under `/sources`.
- Frontend: the X connector card gains a "Sync bookmarks" action (new `twitter` connector kind with two add buttons); source rows label as "X bookmarks (@handle)".

## Tests
`test_twitter_bookmarks.py` — indexer archives with rendered content + embed_stale, un-bookmarked posts stay archived, add-source resolves the account id and names the source. Map-consistency suite (`test_integration_sources.py`) passes unmodified. Frontend integration component tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #799, opened 2026-07-13
Instagram saves: extension push + ScrapeCreators hydration + media archive

samzliu:
Stacked on #798. Fifth backend slice of the commonplace-book feature.

## What
- `POST /api/v1/me/saved-items` `{platform: "instagram", items: [{url, saved_at?}]}` — the extension (logged into instagram.com) pushes the user's saved-post URLs. The endpoint **get-or-creates** the provider-less `instagram_saves` source (no setup ordering), inserts pending skeleton rows keyed by shortcode, and kicks an immediate sync. Unparseable URLs are a loud 400.
- Hydration indexer (`backend/integrations/social_saves/indexer.py`): per pending row, ScrapeCreators post details (`/v1/instagram/post`) + reel transcript (`/v2/instagram/media/transcript`) with the product-level `SCRAPECREATORS_API_KEY`, **plus the media bytes themselves into object storage** (50MB cap) — true archive: the save survives post deletion / private accounts. Per-item failures land on the row with attempts capped at 3; sync interval (30 min) is the retry pass, freshness comes from pushes.
- Reads of unhydrated rows fail loud with 409/422 + the recorded error (same contract as `drive_documents`); hydrated docs return a fresh presigned `media_url` that the source-doc viewer renders as inline `<video>`/`` — the one viewer change in the feature.
- `test_provider_disconnect_cleanup_mapping_covers_registered_providers` relaxed once (`<=`) for provider-less groupings, with a comment.

## Config
`SCRAPECREATORS_API_KEY` env var, needed on **web and worker** (env-parity).

## Tests
`test_instagram_saves.py` — push get-or-create + dedupe + sync kick, loud URL rejection, full hydration (content/transcript/media/presigned read), failure-on-row + fail-loud read. Map-consistency suite passes.


samzliu:

PR #801, opened 2026-07-13
Extension: web clipper, clip-all-tabs, and bookmarks import

samzliu:
Stacked on #800; talks to the backend endpoints from #795–#797.

## What
- **Clip this page**: context menu "Save page to Stash" + popup button → `clipper.ts` injected on demand (activeTab grant only — no `<all_urls>`) captures the live, logged-in DOM → `POST /me/clips/page`. YouTube/arXiv URLs return 202 and finish server-side (transcript / paper PDF).
- **PDF dispatch**: Chrome's PDF viewer refuses script injection, so those tabs take the byte-fetch path (`credentials: 'include'`, works on auth-gated PDFs) → `POST /me/clips/file`.
- **Clip all open tabs**: popup button sends every http(s) tab URL to `POST /me/imports/tabs`; popup polls batch progress.
- **Bookmarks import**: popup file picker uploads a bookmarks.html export to `POST /me/imports/bookmarks`, same progress view — no web-app UI needed.
- Fail-loud UX: clip failure → red `!` badge + Chrome notification with the real server error + `lastError` in the popup; success → transient green `✓` badge + "view clip" link.

## Verified
`npm run typecheck` + `npm run build` clean; `dist/clipper.js` is injected via `chrome.scripting.executeScript` (no `web_accessible_resources` needed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #802, opened 2026-07-13
Extension: daily tab-less ChatGPT/Claude poll

samzliu:
Stacked on #801. No new backend surface — reuses the existing `syncConversation` upload path (hash dedupe makes re-uploads cheap).

## What
- `chrome.alarms` daily tick → list the 50 most recent conversations per platform (**ChatGPT**: `/api/auth/session` → `/backend-api/conversations?order=updated`; **Claude**: `/api/organizations` → `/chat_conversations`) as cookie-attached service-worker fetches, then extract + sync anything updated since the last successful pass — the exact same `chat_apis` client the content scripts use, just with an absolute origin and `credentials: 'include'`.
- Watermark (`lastChatPollAt`) advances only on a fully successful pass; failure of either platform sets the badge + popup error and retries at the next alarm. No notifications — a logged-out site would nag daily.

## Manual check
Fire the alarm from the SW console: `chrome.alarms.create('chat-poll', {when: Date.now()})`.

### DIFF
--- a/chrome_extension/src/background.ts
+++ b/chrome_extension/src/background.ts
@@ -4,6 +4,7 @@
 // rest as replace-mode transcripts so a growing chat keeps updating the
 // same Stash session.
 
+import { initChatPoll } from './background/chat_poll';
 import {
   clipActiveTab,
   clipAllTabs,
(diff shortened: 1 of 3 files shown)


samzliu:

PR #803, opened 2026-07-13
Extension: Instagram saves capture with daily auto-visit

samzliu:
Stacked on #802; talks to `POST /me/saved-items` from #799. Completes the v1 commonplace-book stack.

## What
- `src/content/instagram.ts`: on instagram.com, reads the saved-posts list via Instagram's own same-origin API (`/api/v1/feed/saved/posts/`, `x-ig-app-id` + csrftoken from cookies — the proven pattern saved-posts exporter extensions use). Paginates up to 200 items; only **URLs** leave the page (ScrapeCreators hydrates content server-side).
- 24h throttle owned by the background (`SHOULD_FETCH_SAVES` → single writer of `igFetchedAt`); empty harvests still record the pass.
- **Daily auto-visit** (user-requested): an alarm opens a non-active tab to instagram.com when a harvest is due; the tab closes when `SAVED_ITEMS` arrives or after a 2-minute timeout, which surfaces loudly ("are you signed in to instagram.com?") on the badge.
- Failures (logged out, API drift) report through `SAVED_ITEMS_FAILED` → badge + popup error.
- Manifest: instagram.com content script + host permission.

## CWS note
Listing must disclose that the extension reads the user's own Instagram saves and sends them to the user's Stash account (Limited Use). Reading the saved list via the internal API is the same user-owned-data pattern as existing saved-posts exporters; accepted startup risk per plan.

## Verified
`npm run typecheck` + `npm run build` clean. Endpoint shape verified against a logged-in session should happen before store submission (internal APIs drift).

### DIFF
--- a/chrome_extension/esbuild.config.js
+++ b/chrome_extension/esbuild.config.js
@@ -6,6 +6,7 @@ const options = {
     chatgpt: 'src/content/chatgpt.ts',
     claude: 'src/content/claude.ts',
     clipper: 'src/content/clipper.ts',
+    instagram: 'src/content/instagram.ts',
     popup: 'src/popup/popup.ts',
   },
   bundle: true,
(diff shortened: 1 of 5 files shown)


samzliu:

PR #825, opened 2026-07-16
Heavi rules-of-the-road connector: live VFS reads, indexed for search

henry-dowling:
## What

A new **Heavi** integration that mounts a customer's "rules of the road" (their `learnings` table — per-customer-org part-sourcing preferences) into Stash as a connected source: one folder per customer org, one small text file per rule.

Design principle agreed with the customer: **Heavi's Postgres stays the ground source of truth.** Stash never writes back, never regenerates rules, and never serves a copy — every read is live.

## How

- **First `api_key` provider** — exercises the pasted-credentials flow that existed in `base.py`/`router.py` but had no user until now. Connect = paste endpoint URL + bearer token; validated by fetching the learnings once; stored Fernet-encrypted as a JSON bundle (the Gong pattern).
- **Live `ls`/`cat` — the only read path.** VFS reads under the source fetch the customer's endpoint on every call. A rule added or deleted upstream shows on the very next `ls`, no sync needed. `cat` resolves full paths *and* raw rule ids. If the endpoint is down, the read fails loud.
- **No local copy (deferred).** Whether rules get a search index at all is parked until the unified-search rework (#860) settles — no indexer, no sync schedule, not in `CONTENT_TABLES`. The `heavi_learning_docs` table (migration 0158) ships **empty** so the shared per-source codepaths (sidebar tree, item counts, doc listing) run against zero rows instead of forking on a table-less source type. Until the #860 decision, rules are browse-only and absent from search.
- **Gated to Heavi.** The provider declares `allowed_email_domains=("heaviai.com",)`; the router hides it from `GET /integrations` and 403s connects for anyone without a verified `@heaviai.com` email or the Heavi workspace scope. Frontend visibility is server-driven (no card render before the provider list loads, no flash).
- **Frontend** — connector card, icon, labels. The credential form itself was already generic.

## Endpoint contract (Heavi side, separate PR in their repo)

`GET <base_url>` + `Authorization: Bearer <token>` → bare JSON array of `{id, org, summary, source_type, source_id?, created_at, updated_at?}` — their existing `GET /api/learnings` serialization plus a required `org` field (the customer org's display name; rules render under per-org folders, so a row without it is rejected). Returns all orgs' rules. Until they ship it, the connector has nothing real to point at.

## Verification

- Backend: full suite green locally, 13 heavi tests (rendering/org-nesting, client shape validation fail-loud, provider, gating, and an API-level E2E: register → connect → add source → live ls → live cat by path and raw id → upstream mutation visible on next ls with no sync). A consistency test pins the live-read wiring (table present, deliberately no indexer, no sync schedule). Frontend: 210 pass; ruff + eslint clean.
- Full browser walkthrough against a local HTTPS stub: connected via settings, added the source, **Browse listed every rule while the source showed "never synced · 0 items"** — live reads proven in the product. Screenshots in the comments (early ones predate the per-org folders and the index deferral; see the comment thread for the design evolution).
- Known-red CI check: `test_github_index_success_logs_internal_source_id_only` is broken on `main` itself (#858 added an unmocked network call); unrelated to this PR.


henry-dowling:
## Screenshots

**Connect form** — first use of the api_key credential flow (endpoint URL + bearer token):

Heavi credential form

**Connected + source added** — external_ref is the customer's endpoint URL:

Heavi source added

**The live-read proof** — Browse lists all 3 rules while the source still shows "never synced · 0 items", so the listing can only be coming from the endpoint, not the index:

Live browse with zero indexed items

(Taken against a local HTTPS stub implementing the endpoint contract; screenshots hosted in a gist to keep binary assets out of the repo.)

henry-dowling:
## Update: provider gated to Heavi accounts

The Heavi card is now invisible and unconnectable for everyone except Heavi:

- The provider declares `allowed_email_domains = ("heaviai.com",)`; a new router-level gate (`_user_may_use_provider`) filters it out of `GET /integrations` and returns **403** from both connect endpoints for anyone else.
- "Heavi account" = a **verified** `@heaviai.com` email or the **Heavi workspace's scope user** (`workspaces.domain`) — the same `email_verified` trust anchor as derived workspace membership, so an unverified `fake@heaviai.com` signup sees nothing.
- Frontend visibility is server-driven: the settings connector list and the explorer Tools section only render providers the backend returned, so no client-side knowledge of the rule.
- Providers without `allowed_email_domains` are untouched — open to everyone as before.

Tests: 891 backend (3 new: hidden+403 for outsiders and wrong-domain emails, visible for verified heavi emails, unverified-revoked, workspace scope user allowed), 208 frontend.

henry-dowling:
## Update: closed the render-flash gap

Before this fix, the connector list rendered all cards from the static client-side list and *then* hid server-omitted ones — so every user would have seen the Heavi card flash for a few hundred ms before it vanished. Now no card renders until the server's provider list has loaded, and a direct `/integrations/heavi` URL shows "Unknown integration" for non-Heavi users.

Verified live against a local stack: `GET /integrations` for a user with no email → no `heavi`; same user with a verified `@heaviai.com` email → `heavi` present; verification revoked → hidden again.

henry-dowling:
## Update: per-org folders

Rules are per-customer-org in Heavi's schema (`learnings.organization_id`), so the mount now renders **one folder per org**:

```
/sources/Heavi — Rules of the Road/
  KingFleet/
    Avoid Aftermarket brand brake calipers… (learning_…x7f2ab)
  Kleyn Mobile/
    Never use generic sensors for DEF systems… (learning_…q8r4ef)
```

**Endpoint contract change (before Heavi builds it):** each row now requires an `org` field (the customer org's display name) — `{id, org, summary, source_type, source_id?, created_at, updated_at?}` — and the endpoint returns all orgs' rules. A row without `org` is rejected (unattributable). This matches how the rest of their stash is already organized per org (session folders, `/memory/org/<name>`).

No frontend changes needed — the browse UI already nests path segments into folders.

henry-dowling:
## Update: search index deferred until #860

The local copy of the rules is gone from this PR — no indexer, no sync schedule, not in `CONTENT_TABLES`. Rules are now **live-read only**: browse/`ls`/`cat` hit the customer endpoint (as before), and they simply don't appear in search or embeddings yet.

Rationale: #860 (unified full-text search) live-fetches federated sources' text at query time and ranks it alongside indexed content — exactly the no-copy search pattern that would fit this source (tiny corpus, one cheap GET). Rather than ship a 6h-stale copy now and rework it after #860 lands, the indexing decision waits for that machinery. The `heavi_learning_docs` table ships empty so the shared per-source codepaths (tree, item count) run against zero rows instead of forking on a table-less type; `indexer.py` → `render.py` (pure rendering for the live path).

Net: −65 lines, no staleness window to explain, and one less semantic collision with #860 (this PR no longer registers anything in the search maps it reworks).

PR #839, opened 2026-07-17
Reject conversational LLM output in session title generation

henry-dowling:
Follow-up to #837, which cleaned up the ~1,400 bad cached titles. This fixes the generation path so it can't mint new ones.

## Root cause

`_generate_title` handed the model a raw transcript as the user message. The model sometimes **replied to the conversation instead of titling it** — that's where "You're right—I apologize for the confusion" came from — or refused ("I need more context to generate a title"). Whatever came back was cached verbatim, and the source-hash staleness check then froze it forever.

## Changes

- **Prompt**: the source is now wrapped in `<transcript>` tags and the system prompt says explicitly it is titling, not chatting, and must never ask for more context.
- **Output validation**: `clean_generated_title` rejects reply-shaped and refusal-shaped output (`You're …`, `I'll …`, `I need …`, `Yes—…`, bare `Perfect!`), so a conversational response can never be cached as a title. Covered by new unit tests, including a guard that task-shaped titles ("Fix session title generation") still pass.
- **No retry storm**: when output is rejected, the task caches the deterministic event-derived title (same derivation the listing fallback uses) instead of storing nothing — otherwise the session would be re-sent to the LLM by `reconcile_missing` every 60s forever.

## Testing

- 35 title/transcript/rename tests pass against a fresh Postgres (which also re-validated migration 0154 from #837 via the conftest `alembic upgrade head`).

### DIFF
--- a/backend/migrations/versions/0155_regen_reminted_garbage_titles.py
+++ b/backend/migrations/versions/0155_regen_reminted_garbage_titles.py
@@ -1,0 +1,35 @@
+"""Repeat the 0154 garbage-title sweep for rows re-minted by the old prompt.
+
+0154 deleted the legacy garbage titles, but the regeneration backlog was
+processed by the old prompt (this deploy is what fixes it), which re-minted
+reply-shaped titles for ~60% of them. Same delete, same patterns: this runs
+at boot before the fixed pipeline starts, so everything regenerates clean.
+
+Revision ID: 0155
+Revises: 0154
+Create Date: 2026-07-16
+"""
+
+from alembic import op
+
+revision = "0155"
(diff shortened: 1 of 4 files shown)


henry-dowling:
Added migration 0155: the 0154 backlog is regenerating under the old prompt right now and has re-minted garbage for ~60% of processed rows (167 of ~280 as of 03:10Z). 0155 repeats the same delete at boot — since migrations run before the new code serves, one deploy of this PR wipes the re-minted rows and regenerates everything through the fixed pipeline. Recommend merging soon to stop burning Haiku calls on titles that will be deleted again.

PR #846, opened 2026-07-17
X saves: kind folders in the VFS + a real tweet viewer

samzliu:
## What

Saved tweets now browse as **folders** in the VFS and open in a **real viewer** instead of a wall of monospace text.

- **Kind folders.** Every save lands under its kind: `Bookmarks/`, `Posts/`, `Replies/`, `Articles/` (path = `<Folder>/<tweet id>`). The `x_saves` source is now `navigable`, so the integration page shows folders you can drill into and filter by kind. Migration `0156` renames existing flat rows and flips the capability in one shot.
- **Browse previews.** Each row previews the copied tweet text; an un-hydrated save (still just its numeric id) shows *Loading…* until the worker fills it in.
- **Inline tweet viewer.** Clicking a save expands it in place: the tweet renders as prose with a muted byline / reply-context / link line, and archived images/video play above it.

## Also in here (worker health)

- **Hydrate in the same pass.** Backfilled tweet ids are enqueued *before* the hydration batch runs, so a newly-saved tweet fills in this sync instead of waiting for the next one.
- **GitHub PR fetch follows redirects.** Renamed repos `301` their PR API path; without following it every PR under that repo raised and flooded the Celery worker, starving X syncs. `follow_redirects=True` fixes it.

## Testing

- Backend: `pytest` — **853 passed** (fresh DB through migration 0156).
- Frontend: `npm run lint` clean on the touched page, `npm test` — **209 passed**.
- Extension: `npm run build` clean.

Screenshots of the folder tree + viewer to follow in a comment.

### DIFF
--- a/backend/integrations/x_saves/indexer.py
+++ b/backend/integrations/x_saves/indexer.py
@@ -46,6 +46,18 @@ def tweet_url(tweet_id: str) -> str:
     return f"https://x.com/i/status/{tweet_id}"
 
 
+# Saves are foldered by kind in the VFS: path = "<Folder>/<tweet id>".
+KIND_FOLDER = {"Bookmark": "Bookmarks", "Post": "Posts", "Reply": "Replies", "Article": "Articles"}
+
+
+def save_path(kind: str, tweet_id: str) -> str:
+    return f"{KIND_FOLDER.get(kind, 'Other')}/{tweet_id}"
+
+
+def _tweet_id_from_path(path: str) -> str:
+    return path.rsplit("/", 1)[-1]
+
+
 async def index_x_saves(source: dict) -> str | None:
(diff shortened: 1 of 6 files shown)


samzliu:
### Product screenshots

**Browsing X as folders** — bookmarks, posts, replies and articles each get their own folder in the VFS:

X saves folders

**Tweet-text previews + inline viewer** — each save previews the copied tweet (an un-hydrated save shows *Loading…*); clicking one expands it as prose with a muted byline/link, not a monospace blob:

Bookmarks with the inline tweet viewer

**Replies keep their thread context** — the viewer shows the "In reply to @…" line the hydrator captured:

Reply with in-reply-to context

<sub>Captured on this branch against a locally-seeded X source (the folders UI isn't on prod yet).</sub>

PR #852, opened 2026-07-17
Hide the Gong connector card until an OAuth app exists

henry-dowling:
we should actually make gong work at some point! but not until someone asks us for it!

## Why

Reviewing prod's Render env today showed there are no `GONG_OAUTH_*` credentials anywhere — the Gong OAuth app was never created. So the Gong card on /integrations has only ever produced "Gong OAuth is not configured for this server" for every user who clicked Connect, locally and in prod.

## Change

Comment out (not delete) the Gong entry in the connector catalog (`connectors.tsx`) — the single spot that renders the card. The provider implementation stays intact and untouched; re-enabling is uncommenting one block once someone registers the OAuth app with Gong and lands the creds in Render + `.env`. A comment at the site says exactly that.

Everything else Gong-related (backend provider, icons, `gong_calls` render paths) is unreachable without the card and left alone — existing-source render paths would still work if any sources existed (none can, since connecting was never possible).

## Verified

- `npm run lint`: 0 errors (3 pre-existing warnings in unrelated files).
- `npm test`: 209/209 pass.

### DIFF
--- a/frontend/src/components/integrations/connectors.tsx
+++ b/frontend/src/components/integrations/connectors.tsx
@@ -102,13 +102,16 @@ export const CONNECTORS: Connector[] = [
     kind: "auto",
     blurb: "Meeting notes and transcripts.",
   },
-  {
-    provider: "gong",
-    label: "Gong",
-    sourceType: "gong_calls",
-    kind: "auto",
-    blurb: "Call transcripts, kept in sync.",
-  },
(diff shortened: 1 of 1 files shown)


henry-dowling:
Visual delta: the Gong card (between Granola and PostHog on /integrations) disappears; no other UI changes. Skipping the usual before/after screenshots for now — the fixed dev ports are occupied by a live session, and the change is a pure removal of one catalog entry. Can attach a screenshot from a local run after merge if wanted.

PR #857, opened 2026-07-17
Skip repo files too big to full-text index instead of failing the sync

henry-dowling:
@ColemanDuPlessie - heads up, we had some github files that were so big that we couldn't fit them into a `tsvector`- when you get a chance, could you think of ways to work around this? seems like we're probably doing something dumb here. Like, maybe there's another way to index these files...

## Why

`github_documents` has a GIN full-text index on `content`, and Postgres hard-caps a single row's tsvector at **1MB** (non-configurable). The repo crawler skips binaries and files over 10MB, but a *text* file between ~1MB and 10MB sails through and blows up at insert time:

```
asyncpg.exceptions.ProgramLimitExceededError: string is too long for tsvector (1253392 bytes, max 1048575 bytes)
```

That exception aborts the **entire repo's sync** — none of the repo's files are searchable, not just the big one. And since failed syncs never store a cursor, the #855 HEAD-unchanged skip never engages: the repo re-downloads its full zipball every cycle just to fail again. Two prod repos are stuck in exactly this loop (`dopeboy/trycadbury.com`, `vadymkostucok/cadbury`).

## What

Text files over **512KB** are now skipped during the crawl, exactly like binaries and >10MB files — same skip rule, one codepath. Oversized text files are near-always minified bundles, lockfiles, or data dumps with little search value. Skipping (rather than truncating stored content) means the agent never reads a silently-chopped file when browsing a repo. The stuck repos will sync clean on their next cycle, no migration needed.

## Tests

New: a repo containing an oversized text file syncs successfully and indexes everything else. Ruff clean, sync-related suites pass (113 tests).

### DIFF
--- a/backend/integrations/github/indexer.py
+++ b/backend/integrations/github/indexer.py
@@ -26,13 +26,19 @@ from ..storage import get_valid_token
 from .archive import UnsupportedHostError, _parse_owner_repo, resolve_archive_url
 from .importers.repo import (
     MAX_FILES,
-    MAX_PER_FILE_BYTES,
     _download_archive,
     _is_skipped_path,
     _strip_top_level_prefix,
 )
 
 logger = logging.getLogger(__name__)
+
+# Postgres caps a row's tsvector at 1MB, and github_documents carries a
+# full-text index on content — a bigger file aborts the whole repo's sync
+# at insert time. Text files over this cap (minified bundles, lockfiles,
+# data dumps) are skipped like binaries; 512KB keeps the vector safely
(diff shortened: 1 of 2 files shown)


henry-dowling:
this was the errant file, btw https://raw.githubusercontent.com/dopeboy/trycadbury.com/main/src/assets/logoColor.svg

henry-dowling:
a logo from our friend manish's app https://www.linkedin.com/in/manish-sinha-9675b366/

PR #860, opened 2026-07-17
integrations: unified full-text search across all integrations

ColemanDuPlessie:
Search now runs over all connected integrations, not just stash. To facilitate this, we grab the full text of each external document our federated search returns (truncated at 20K chars for now, could be adjusted), and compare them using the same Postgres we were using before for within-Stash search.

This change unifies the CLI search command and the browser-based search bar. As before, the search will display a snippet with the first occurrence of the searched word or phrase. Newly, the CLI will display the same snippet (a few hundred chars) when agents use it to search.

### Remaining/new issues
- Google drive doesn't search over body text when searching over all of drive rather than a linked folder (this is because it would both take extra per-document API calls, and is sometimes impossible, such as when the document being returned is an image or binary file)
- Trimming to 20K characters could be problematic and give wrong results if searched text is near the end of very long documents. However, we do need some way of handling massive files (e.g. full meeting transcripts)
- We now search Twitter on every search call; this could become problematic with their API free tier, so may want to have Twitter be off-by-default.
- Still need to implement CLI flags/browser search options to include/exclude individual integrations (behavior will be everything [or everything-but-Twitter] on by default).
- Due to inconsistent data from integrations, not all search results have a "date modified" field, so I've tentatively dropped that from search result cards; could potentially add it back for Stash results only (or Stash + relevant integrations only) if needed.

### One merged, ranked, paginated result list
- `GET /api/v1/me/sources/search` now merges all sub-searches (sessions + pages + federated sources) onto a single `ts_rank` relevance scale (`_uniform_ranks` re-scores every hit's `name\nsnippet` in one query, since per-source orderings aren't comparable).
- Response shape is now an envelope: `{"results": [...], "has_more": bool}`, with `offset` (default 0) and validated `limit` (1–100). A sentinel row past the page proves `has_more`.
- Error/truncation markers trail every page and are not counted toward pagination.
- Twitter/X now fans out in unscoped searches like every other source — the `SCOPED_ONLY_SEARCH_TYPES` carve-out is gone.

### Search hits carry full text
- FTS snippets over indexed content are now capped at 20,000 chars (`SEARCH_SNIPPET_CHARS`) instead of 400, so callers can rank on real document text.
- Live provider searches fetch full bodies: Gmail renders the full message (not Gmail's ~150-char snippet), Twitter prefers `note_tweet` long-form text, Jira/Asana request full field sets and render the issue/task, Linear's GraphQL search returns complete `LinearIssue` objects.
- Ask-the-stash keeps model context small: `_trim_snippet` windows each hit to ~400 chars around the first query match before handing results to the agent.

### Frontend
- The search page is rewritten around the single unified endpoint (replacing separate `searchEvents`/`searchPages` calls). Connected-source hits render inline via the new shared `SourceDocViewer` component (extracted from the integrations provider page), with an "Open in {provider} ↗" deep link.
- Marker rows surface as ⚠ notices (dead source, "showing first N of ~M"), and `has_more` renders a refine-your-query hint.

### CLI / MCP
- `stash search` prints a 300-char head per hit and warns when more matches exist; `--json` carries the full text. The MCP `stash_search` tool gains `offset` and documents the envelope.

## Tests
- Uniform-scale merge, pagination, markers-trail-every-page, rank-zero retention, negative-offset 422, capped-full-text snippets, Gmail/Jira/Asana/Twitter full-field fetches, `_trim_snippet` windowing, and CLI envelope handling. (148 tests across the touched files pass; `ruff check` + `format` clean.)

## Known issue (pre-existing on this branch)
`index_gmail` fetches up to 100 message metadata requests in one unbounded `asyncio.gather`, which trips Gmail's per-user quota (429) and fails the sync every run. Needs a concurrency cap (e.g. `asyncio.Semaphore`) — left for a follow-up.


henry-dowling (changes requested):
Nice work! This is not an easy task and this is a good first pass at an implementation.

Some comments:

(0) I found a frontend bug! Try clicking "open" on any integration and observe that there's a broken type: https://www.loom.com/share/7be7868da8bc4f1abb5a9997f1f80316?from_recorder=1&focus_title=1

This is mostly on me, actually. I put up #865 to prevent this kind of thing from happening again.

(1) Re: one of the issues you brought up around the ranker only considering the first 20k chars of a doc, I have an idea: why not take the 20k character window around the *first relevant part of the document that caused it to be a match* rather than just the *first* 20k?

(2) I noticed that for our *federated sources*, snippets are not capped at 20k chars, only indexed sources. We should add a 20k char cap for consistency. You actually explicitly mention that you do this in your PR, so clearly you had the right intentions but as it stands this is not implemented.

(3) The way we do pagination right now is pretty jank. Like, to get results 21-40 from a query where we paginate with page size 20, we actually re-run the entire search again, and then clip results to [21, 40]. This has the unfortunate side effect that a result could conceivably appear twice (suppose it shifts from 20 to 21) or never (21 to 20) across two searches. After considering a few approaches (stateful pagination, always fetching first 3 * page_size, etc) the best solution I can come up with is to just not allow pagination at all, full stop. It's too hard to have both pagination and re-ranking at runtime. What we could do instead is just allow the LLM to ask for a desired num_results. Do you have any better ideas here?

(4) I noticed something interesting about the way we handle "exact ID is a match but that ID shows up nowhere in the text"- we rank that *below* everything else. IMO, an exact provider ID match is an extremely relevant result. Why not put that first?

(5) Our MCP and CLI are inconsistent in their enforcement of truncation in their results. CLI returns the first 300 chars of a result whereas the MCP implmentation, `stash_search`, returns the whole thing, so potentially 20k chars per result. Can you please change the MCP policy to match that of the CLI? Also, I think we should return a 300 char window around the relevant text, not the first 300 chars (like your onboarding project!)

(6) Re: "Due to inconsistent data from integrations"- what if we just made date_modified an optional field?

(7) Small comment FYI: there is an opportunity for an edge case where the provider and the ranker use a different tokenizer, causing our ranker to completely miss the reason that a result is relevant. For example, if Gmail's automatically translates text, but ours does not.

(8) There is a difference in which text is considered for computing a match by our integration providers versus which text is considered for computing a rank by our postgres ranker. Can you please fix the low-hanging fruit here so that we rank on as close to the same text as the text that we match on as possible? Below is a Fable summary

<fable>
- Gmail: Cc recipients (~1 line). The search path fetches format=full, so every header is in hand; _render_message renders From/To/Date and skips Cc. "Emails involving this person" is a bread-and-butter agent query, and cc'd mail currently ranks at zero.
- Gmail: attachment filenames (~5 lines). The message parts we already walk in _extract_body each carry a filename we ignore. One Attachments: pricing.xlsx, deck.pdf line vindicates the entire matched-via-attachment class of hits — the biggest and most common gap.
- Jira: the environment field (~2 lines). JQL's text ~ searches it; it's just missing from the requested-fields tuple. Jira's render is otherwise the most complete of the bunch.
- Linear: comments (~10 lines). It's GraphQL, so adding comments { nodes { body } } is the same single request with a bigger response — unlike Asana, where comments would cost an extra API call per hit. 
</fable>


(9) Small thing: Can we remove "twitter" from **FEDERATED_SEARCH_TYPES**? And delete `SCOPED_ONLY_SEARCH_TYPES`?

henry-dowling (approved):
Send it!

Issue #780, opened 2026-07-11
Image version doesnt exists & Error on memory curator


dickyfarhansyah:
### Component

Self-hosting / Docker

### What happened?

Tried deploying it locally, After cloning the repo and starting the compose it raised an error saying version 0.1.226 not found, verifying to the GHCR version and its correct no 0.1.226 exists the latest is .224 hence i fix it manually by changing the version to 0.1.224, then its starting fine.

---

Memory curator raised an error when triggered manually saying
`agent turn failed: [Errno 2] No such file or directory: 'claude'`

Backend logs looks clear but the worker logs has some errors.

### What did you expect?

Memory curator works fine and i dont need to manually define image version (tho its fine for me to edit the compose)

### Steps to reproduce

1. Follow the self-host steps
2. Set the EMBEDDING_MODEL to `nvidia/llama-nemotron-embed-vl-1b-v2`
3. Set the EMBEDDINGS_DIM to 2048 (to match the embeddings model output dims)
4. Follow the setup and usage steps
5. Run the memory curator manually

### Environment

- Stash version: 0.1.224
- Distro: Ubuntu 24.04.4 LTS
- Docker: 29.6.1
- Model provider: Openrouter

### Relevant logs or error output

```shell
worker-1  | 2026-07-11T16:55:22.409616065Z [2026-07-11 16:55:22,407: INFO/MainProcess] Task backend.tasks.agent_schedules.run_curator_now[2927d662-0ba1-47b6-a275-3fd2c3bd3f08] received
worker-1  | 2026-07-11T16:55:22.578440510Z [2026-07-11 16:55:22,575: ERROR/ForkPoolWorker-4] Task backend.tasks.agent_schedules.run_curator_now[2927d662-0ba1-47b6-a275-3fd2c3bd3f08] raised unexpected: RuntimeError("agent turn failed: [Errno 2] No such file or directory: 'claude'")
worker-1  | 2026-07-11T16:55:22.578492068Z Traceback (most recent call last):
worker-1  | 2026-07-11T16:55:22.578497928Z   File "/app/backend/services/sprite_agent_service.py", line 469, in run_chat
worker-1  | 2026-07-11T16:55:22.578502810Z     async for event in _turn_events(
worker-1  | 2026-07-11T16:55:22.578507035Z   File "/app/backend/services/sprite_agent_service.py", line 238, in _turn_events
[...]

PR #861, opened 2026-07-19
fix: fail loud with a clear message when the harness CLI is missing from the worker

JoeJoeflyn:
## Summary

Fail loud with a clear, actionable error when the harness CLI (claude/codex/opencode) is missing from the worker host, instead of the cryptic FileNotFoundError from issue #780.

## Changes

- `backend/services/sprite_agent_service.py`: check `shutil.which(auth.harness.bin)` before spawning the subprocess in `run_chat`; raise `RuntimeError` naming the missing binary if absent
- `backend/tests/test_curator.py`: added `test_missing_cli_raises_clear_error` — mocks `shutil.which` to return `None`, asserts the error names `claude` and mentions `PATH`

## Related issues

References #780

## Test plan

- [x] Existing tests pass (`python -m pytest backend/tests/test_curator.py -v` — 22 passed)
- [x] New tests added (`test_missing_cli_raises_clear_error`)
- [x] Tested manually — started the self-hosting stack via `docker-compose.prod.yml`, confirmed `claude` is absent from the worker container, reproduced the original `FileNotFoundError`, then confirmed the new check raises a clear `RuntimeError` naming the binary

### DIFF
--- a/backend/services/sprite_agent_service.py
+++ b/backend/services/sprite_agent_service.py
@@ -21,6 +21,7 @@ import json
 import logging
 import re
 import secrets
+import shutil
 from collections.abc import AsyncIterator
 from uuid import UUID
 
@@ -594,6 +595,13 @@ async def run_chat(
         raise NeedsAuth
     except agent_auth.ProviderNotConfigured:
         raise RuntimeError("cloud agent is not configured")
+    # Fail loud before spawning: a missing CLI produces a cryptic FileNotFoundError
+    # deep in asyncio that doesn't tell the operator what to install.
+    if not shutil.which(auth.harness.bin):
+        raise RuntimeError(
+            f"agent CLI '{auth.harness.bin}' not found on PATH — "
+            f"install it on the worker host so the agent can run turns"
+        )
     async with _TurnLock(session_id):
         history = await _load_history(owner_user_id, session_id, user_id)
         await memory_service.push_event(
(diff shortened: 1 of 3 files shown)


henry-dowling:
Thanks for the PR! Will take a look soon

ColemanDuPlessie (commented):
Looks good! One note that the codebase previously didn't use the shutil python library in backend/, only in the CLI and plugins. It's probably fine, but perhaps something to be aware of.

JoeJoeflyn:
My bad! I didn't notice `backend/` wasn't using `shutil` before. Do you want me to remove it and use a different approach instead?

ColemanDuPlessie:
That's alright! Given that we already use shutil elsewhere in stash, it's okay to also introduce it into backend/

It seems like this change is causing a couple of the backend tests to fail (there are 6 failures in the CI, and all but test_integration_indexer_log_security.py::test_github_index_success_logs_internal_source_id_only come from this change): it looks like they're all coming from the fact that our pre-existing tests trip up your code and cause it to fail loud when it doesn't recognize the tests' mocks. Other than that, this PR looks good, so once that's fixed, we can merge it!

JoeJoeflyn:
Thanks, really appreciate it! Glad I could help improve this. Happy to contribute!

henry-dowling:
Hey @JoeJoeflyn do you plan on addressing the test failures here? Lmk otherwise I'm gonna close since this is getting stale

JoeJoeflyn:
yep, i forgot this

PR #884, opened 2026-07-23
Self-serve agent platform: 7-harness coverage, synced skills, MCP registry, workspace scope

samzliu:
## Motivation

The self-serve audit (session-linked below) found three product gaps: the Claude Code plugin only worked for one person in one harness, skills had no install lifecycle, and "team" appeared nowhere except copy that dead-ended. This PR ships the whole plan.

### 1. Harness coverage (7 agents)
- `stash signin` auto-installs the plugin for **Gemini CLI**, **Openclaw**, and a new **Hermes** adapter (on top of claude/cursor/codex/opencode). Assets vendored into the `stashai` package; byte-parity enforced by `test_assets_in_sync`.
- Openclaw's code scanner blocks `child_process` plugins, so the installer passes `--dangerously-force-unsafe-install` — spawning the Python hook scripts *is* the extension's mechanism. Live-verified: extension loads in the gateway.
- Hermes adapter built from the real docs (stdin JSON envelope; `pre/post_llm_call` fire once per user turn; hook stdout must be `{}`). Installer upserts a marker block into `~/.hermes/config.yaml` and **fails loud** if the user owns a top-level `hooks:` key (duplicate YAML keys are last-one-wins).
- Docs de-staled: `stash connect` never authenticated anyone — every surface now says `stash signin`; removed the dead `auto_curate` plugin config and a literal `</content></invoke>` artifact in the README.

### 2. Skills as a synced VFS
- Installs tracked in `~/.stash/installed_skills.json`; new `stash skills uninstall / list --installed / follow / unfollow`.
- The session-start `skills sync` now **auto-updates installed skills** (cloud renames included) and reports via a one-shot next-session notice — the skills dir never changes invisibly.
- `follow` auto-installs skills shared with you; new `GET /me/shared-skills/{folder_id}/contents` lets share recipients materialize **unpublished** skills (the share, not publication, is the grant — tested incl. 403 for strangers).
- Safety: manifest-tracked dirs are excluded from three-way sync, so project mode can never push someone else's skill into your Stash; same-name shadowing conflicts loudly.
- `install_count` (migration 0162) + unauthenticated install ping: views measure curiosity, installs measure adoption.

### 3. MCP tools registry
- `mcp_servers` table (migration 0161; header/env secrets Fernet-encrypted with the integrations keyring), `/api/v1/me/mcp-servers` CRUD, real Tools page (was "coming soon"), `stash tools add|list|remove|install` with clobber-proof `.mcp.json` merging.
[...]

PR #885, opened 2026-07-23
Tools route actually renders the MCP registry page

samzliu:
Post-merge QA of #884 found the new Tools/MCP page never rendered: `WorkspaceShell` shows the tab workbench for every section except `/sessions` and `/memory`, so `/tools` still displayed the workbench ("Open something from the explorer"). The page's vitest passed because it rendered the component directly — the shell never mounted it.

Fix: `/tools` joins the management-page routes, and the route→content decision is extracted into an exported `rendersRouteContent()` with a locking test, so the next unregistered management page fails a test instead of shipping dead.

Verified in the browser after the fix: page renders, add-server flow works end-to-end (form → API → Fernet round-trip → `stash tools install` writes valid `.mcp.json`), remove works. Screenshots on #884's QA comment.

Frontend: 51 files / 220 tests pass, lint clean.

https://claude.ai/code/session_01J5EiY5ADhzVcJTAnsxYUWe

### DIFF
--- a/cli/config.py
+++ b/cli/config.py
@@ -79,6 +79,7 @@ def load_config() -> dict:
 
     if USER_CONFIG_FILE.exists():
         cfg.update(_read_json(USER_CONFIG_FILE))
+        _migrate_legacy_scope(cfg)
 
     if url := os.environ.get("STASH_URL"):
         cfg["base_url"] = url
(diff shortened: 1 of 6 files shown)


samzliu:
Second commit on this branch — also from post-merge QA, ship-fast material:

**Legacy `scope` config values would 400 every request after self-update.** A real machine's `~/.stash/config.json` carried `scope: "repo"` — an old mode string under the key that #884 repurposed for the workspace UUID sent as `X-Stash-Scope`. The backend hard-400s non-UUIDs, so CLI 0.1.296 on such machines fails every call, and the plugin's best-effort hooks **silently stop uploading sessions**. Fix: `load_config` deletes non-UUID scope values once (one-shot format migration, not a fallback), and the plugin client's parse boundary never sends a non-UUID. Four tests lock both sides. Also fixes the stale `stash connect` line in the plugin CLAUDE.md.

CLI+plugins: 273 passed · frontend: 51 files / 220 tests · ruff clean.

PR #886, opened 2026-07-23
fix: make Codex onboarding work on the first try

ColemanDuPlessie:
## Changes

- Codex requires the user to re-trust hooks whenever they change (e.g. from stash's auto-version-bump). As such, instead of running the updater on startup automatically, stash for codex now first asks the user if it should update automatically and stores the answer in a config file.
- For the same reason, removed absolute path to python install directory and replaced it with the new `stash hook run codex` subcommand.
- Changed hooks.json to get rid of the _comment key, which was causing codex to reject it.
- Finally, changed the wizard text to remind the user that hooks in Codex must be explicitly trusted to run.
- Also updated docs & tests

## Context (LLM-written below this point)

Onboarding OpenAI Codex to Stash's transcript hooks failed twice in the field before working (Stash session `Connect Codex agent to Stash transcript hooks--ddbbc749`). This PR fixes every root cause so it works on the first try, without touching Claude Code's onboarding path.

## Bugs fixed

1. **Rejected manifest** — codex `hooks.json` carried a top-level `_comment` key; Codex rejects the entire file for unknown top-level fields. Template is now bare `{"hooks": ...}`, and `_install_codex` strips the stale key from existing installs (one-shot migration).
2. **SessionStart crash** — `uploads_enabled(cfg, event)` vs. the one-arg signature → TypeError on every Codex session start. Fixed in codex; the identical one-liner is also fixed in cursor and opencode (their only change). Claude/Gemini/OpenClaw were already correct.
3. **Trust churn** — installed hooks embedded an absolute path into the versioned python install dir, so upgrades changed the command hash and Codex silently distrusted the hooks. Hooks now call a new stable `stash hook run codex <event>` subcommand — byte-identical on every machine and across every upgrade — and codex's 80-line venv-hunting `_run.sh` is deleted. The installer sweeps stale absolute-path entries while preserving user-added hooks.
4. **Unconditional self-upgrade** — the background `stashai@latest` install at session start (which could swap the env mid-hook and re-trigger #3) is replaced by an ask-once flow: with no preference recorded, the hook emits a `systemMessage` telling the agent to ask the user, recorded via `stash hook auto-update on|off` (`codex_auto_update` in `~/.stash/config.json`). `on` → background upgrade at session start (now trust-safe); `off` → never; manual `stash upgrade` always works.
5. **Silent trust gap** — Codex won't run new/changed command hooks until the user approves them. The wizard now says so instead of claiming streaming is live:

```
  ✓ Codex hook installed  /Users/you/.codex/hooks.json + merged … + …

  Codex hooks were installed or changed — Codex will not run them until you trust them:
    1. Restart Codex.
    2. When Codex prompts to review new hooks (or via its /hooks review), approve the Stash hooks.
  Codex sessions start streaming to Stash only after the hooks are trusted.
```

6. **Stale docs** — README/config.toml.snippet claimed per-repo `.stash`-manifest gating; streaming is actually global.

## Tests

- `plugins/tests/test_codex_hooks_execute.py` — executes all four codex hook scripts end-to-end against fixtures (the exact lines `stash hook run` dispatches). Would have caught both original bugs.
- `cli/tests/test_install_codex.py` — stable hooks.json shape, migration (`_comment` + absolute-path entries swept, user hooks preserved), and reinstall byte-stability (the trust-survival invariant).
- `cli/tests/test_hook_cli.py` — dispatcher error paths, auto-update preference writes, on_stop run via CliRunner.

Full suite: 898 passed. Manual verification: `stash hook run codex <event>` against fixtures — ~0.3s cold start (timeouts are 5–10s), exactly one JSON object on stdout, errors on stderr; migration simulated against a replica of the actual broken hooks.json from the field.

## Rollout note

The new hooks.json only works with a CLI that has `stash hook run`, so users get migrated when they run `stash signin`/`stash settings` with the new CLI — the same binary that writes the hooks serves them, so ordering is self-consistent. Needs a `stashai` version bump to ship to PyPI.

No GUI changes (CLI + hook plumbing only).


henry-dowling (approved):
Great work! My only comment is this: after this PR, we are now in an inconsistent state where Codex has **CLI** own the code for stash uploads, but for Claude Code, the **plugin** owns the code for stash uploads. 

Your change here to make CLI own the code for stash uploads rather than plugin was a good one since that way we can use a consistent python instance for our hooks. I'm glad we were able to get rid of `_run.sh` for Codex, that was really jank.

So- in a follow-up PR, when you get a chance, could you please make a similar change for Claude Code, Gemini, and Opencode? That way our codebase is both consistent and using the more elegant design of having the CLI own upload code that you propose here.

PR #871, opened 2026-07-21
fix: descriptive route prefixes and centralize route paths

JoeJoeflyn:
## Summary
- Rename `/p/` -> `/page/` and `/f/` -> `/file/` for readable URLs
- Move `search`, `settings`, `discover` pages into the `(app)` route group
- Add `routes` object in `workspace-routes.ts` as the single source of truth for app-internal paths
- Add `skillItemPath` helper in `localSkill.ts` for skill-scoped URLs
- Replace 54 hardcoded URL strings with `routes.xxx()` calls across ~20 files
- Replace 17 skill URL strings with `skillItemPath()` calls
- Fix rail `selectSection` nav logic with `ROUTES`/`TOP_LEVEL_ROUTES` constants
- Add "No route magic strings" convention to `CLAUDE.md`

## Stack — merge in order
Split from #863 into 3 PRs (fork → upstream, so all target `main`). **Merge in this order** — each PR includes the previous PR's commit:
1. **#871 (this)** — 1 commit — route prefixes
2. #872 — 2 commits — UI transitions (includes #871)
3. #873 — 3 commits — shadcn UI migration (includes #871 + #872)

After merging #871, GitHub will auto-detect #872 and #873 as having one fewer commit to review. Merge #872 next, then #873.

https://github.com/user-attachments/assets/e8ea8937-bceb-4a75-8376-e71f8777516d

PR #887, opened 2026-07-23
fix: descriptive route prefixes and centralize route paths

JoeJoeflyn:
## Summary
- Rename `/p/` -> `/page/` and `/f/` -> `/file/` for readable URLs
- Move `search`, `settings`, `discover` pages into the `(app)` route group
- Add `routes` object in `workspace-routes.ts` as the single source of truth for app-internal paths
- Add `skillItemPath` helper in `localSkill.ts` for skill-scoped URLs
- Replace 54 hardcoded URL strings with `routes.xxx()` calls across ~20 files
- Replace 17 skill URL strings with `skillItemPath()` calls
- Fix rail `selectSection` nav logic with `ROUTES`/`TOP_LEVEL_ROUTES` constants
- Update backend URL generators (`files.py`, `clip_service.py`, `files_tree_service.py`) to use new `/page/` and `/file/` paths
- Update curator prompt and SKILL.md to reference `/page/` instead of `/p/`
- Update all test assertions for new route paths
- Add "No route magic strings" convention to `CLAUDE.md`

Rebased onto latest `main` (supersedes merged-then-reverted #871).


ColemanDuPlessie:
Hello! We've decided to not go ahead with this PR, as it's mostly a style change: we prioritize open source contributions that clearly improve the stability or functionality of the codebase, whereas this is more of a lateral change in style conventions.

JoeJoeflyn:
@ColemanDuPlessie This isn't a style change; it's a route fix. The only reason it was rolled back is that the backend is still using the old routes with the `/p` and `/f` path prefixes. Once the backend is updated, this change will be necessary. You can check the Discord discussion for more context if needed.

PR #889, opened 2026-07-23
workspace: switch table tabs in place and flush pending layout writes

ColemanDuPlessie:
## Description

Quick PR that fixes the issue where switching to/from a table tab would re-add tabs that were closed <1.2 seconds ago. It does this by first, moving tables into app/(app), and second, ensuring we flush our debounce of tab changes on tab close.

## Problem (LLM-written below this point)

Switching workbench tabs to or from a **table** tore down and rebuilt the entire workspace shell (rail, explorer, tab strip), unlike every other tab kind. It looked like a full page reload, and it resurrected just-closed tabs: close a tab, switch to a table, and the closed tab came back.

## Root cause

- Tables were the one tab kind routed **outside** the `(app)` route group, so `router.replace` on tab focus crossed the group boundary, unmounted the `(app)` layout's `WorkspaceShell`, and mounted a second copy owned by `TableRouteClient`.
- The remount also dropped workspace-layout state: `Persistence` debounces localStorage writes by 1.2s and its unmount cleanup cancelled the pending write, so the fresh mount re-hydrated stale state containing the closed tab.

## Fix

- Move `tables/[tableId]` into the `(app)` route group (URLs unchanged — route groups don't affect paths). The shell was already wired for it: `sectionForPath` matches `/tables/`, `tabFromPath` parses it, and `TabBody` renders `TableClient` embedded. `TableRouteClient` and its duplicate `WorkspaceShell` are deleted; the route body now only renders bare for `?skill=` public views, same shape as `f/[fileId]`.
- `Persistence` now flushes a pending write on unmount and on `pagehide` instead of dropping it, so closes that happen inside the debounce window survive reloads.

## Verification

- 226 vitest tests pass, lint clean, `next build` green.
- Browser QA: tagged the live shell DOM with marker attributes, switched table ↔ page — markers survived both directions (shell never remounted). Closed a tab and reloaded 100ms later — it stayed closed.

## Screenshots

Page tab and table tab now switch in place inside one workbench (same shell, sidebar, and tab strip):

| Page tab active | Table tab active |
| --- | --- |
| page tab | table tab |


henry-dowling (approved):
Nice work, LGTM! And @JoeJoeflyn - kudos to you for inspiring the fix here! We weren't able to merge any of your PRs as-is but the route reorganization you suggested was a great idea.

PR #895, opened 2026-07-24
Move all agent hooks to CLI-owned `stash hook run`

ColemanDuPlessie:
## Summary

Carries the Codex hook-ownership model from #886 forward to every remaining agent. All seven plugins (Claude Code, Cursor, Codex, Gemini, Hermes, OpenClaw, OpenCode) now invoke the stable, machine-independent command `stash hook run <agent> <event>`, which executes the hook scripts shipped inside the stashai package under the package's own Python — one interpreter, no venv hunting, no version skew between scripts and library.

## What changed

**CLI**
- `hook_run` dispatches for all seven agents via a `_HOOK_EVENTS` table (per-agent event lists — e.g. OpenCode has no `on_stop`, Cursor has `on_agent_response`).
- Cursor/Gemini installers sweep stale absolute-path entries from pre-`stash hook run` installs, same as Codex.
- **Fixes `_install_gemini` crashing with a TypeError since #886**, which added a required `markers` argument to `_merge_json_hooks` without updating this call site.

**Per agent**
- **Claude Code**: hooks.json uses the stable commands. Deleted the venv-hunting `_run.sh`, the `stashai-version` pin, `cache_drift.py`, and the ImportError shims in `on_session_start.py` — all existed only because marketplace-cached scripts could lag the library, and that skew window is gone. Scripts are mirrored into `stashai/plugin/assets/claude/scripts/` (scripts only; the manifest version is CI-bumped, so a whole-dir mirror would drift every release). The bump workflow's pin loop is removed — its `git add` glob would have failed once the pin files disappeared.
- **Cursor**: machine-independent hooks.json; manual install is now "copy as-is" instead of `envsubst`.
- **Gemini**: machine-independent settings snippet.
- **Hermes**: no more `PLUGIN_ROOT` substitution — and since Hermes approves hooks per (event, command) pair, the stable command means approval survives stash/python upgrades instead of re-prompting whenever the versioned install path changed.
- **OpenClaw**: `index.ts` spawns `stash hook run openclaw <event>` directly; the `STASH_PYTHON` override existed only for `_run.sh` and is gone.
- **OpenCode**: `plugin.ts` spawns the CLI directly.

**Self-upgrade**
- The `stashai@latest` background upgrade that lived in the deleted `_run.sh` files (and at module top of gemini/hermes session-start scripts) moves to a shared `spawn_self_upgrade()` in `session_upload.py`, called after the uploads-enabled gate. Claude's version pin becomes `@latest` — safe now that scripts and library ship together.
- The deleted cursor/hermes/openclaw/opencode launchers also **fell back to bare `python3`** when venv resolution failed — the silent-fallback failure mode that once uploaded nothing for a month. That path no longer exists.

**Tests**
- The codex-only end-to-end hook-execute test generalizes to every (agent, event) pair across all seven agents (34 script executions against fixtures).
- New tests pin machine-independent commands for claude/cursor/gemini/hermes templates, the Gemini stale-entry sweep, and that the dispatch table exactly matches the shipped script files.

## Notes for review

- OpenClaw's `on_prompt.py`/`on_stop.py` stay dispatchable but `index.ts` never invokes them (messages stream in-process by design) — flagged as follow-up deletion candidates.
- Auto-update policy remains split: Codex keeps its ask-once `codex_auto_update` flow from #886; the other six upgrade unconditionally in the background (pre-existing behavior, preserved).
- Existing Claude Code users pick up the new hooks.json via the marketplace version bump CI performs on merge. Codex users are unaffected (commands unchanged); Hermes users get one final re-approval prompt, after which approvals are permanent.

## Verification

- Full suite: 940 passed; ruff clean.
- Smoke-tested every agent's dispatch live from an isolated HOME: all events exit 0, unknown agents/events fail loudly.


henry-dowling (changes requested):
Some comments

(0) Merge conflicts 

(1) Can we break this up into two PRs? Specifically, can we ship the CLI change first, and *then* the plugin a week later? The reasoning here is this will give people time to update their CLIs (this happens automatically on session start) and install the `stash hook run` CLI verb before the updated plugins attempt to run it.

(2) As a follow-up, can we we give users a more descriptive error message when trying to run an unrecognized hook? This will prevent the need for the above choreograph next time we want to update the way we call the CLI in a hook. Something like: "This stash CLI (0.1.306) doesn't recognize hook agent 'foo'. Your hook config was likely written by a newer stash install. Fix: uv tool install stashai"

(3) Cursor, Gemini, and Hermes: FYI they will break for existing users since the package for these coding agents have no auto-update. Once we update the CLI, these harness's plugins won't be updated to use `stash hook run`. The fix is to re-run `stash signin`. I think we have ~no active users on these, so let's just eat this.

PR #900, opened 2026-07-24
Agents sidebar collapses to agents; scheduled runs become an infinite feed with context resets

samzliu:
## What

**The Agents sidebar is just agents now.** The Recent chats list is removed; each agent row opens that agent's single conversation. Chat agents resume their persistent session. Past ad-hoc chats stay reachable in Sessions.

**Scheduled agents (Memory curator included) get a runs feed instead of fake chats.** Every run is its own session — a fresh context — so the feed renders all runs chronologically with a context-reset separator opening each run (date + `fresh context`, plus `failed` in red when the run's closing message is the stored failure marker). The internal run prompt collapses behind "Show prompt" so the feed reads result-first. A right-hand **Runs rail** maps the feed: one entry per run, click to jump, active entry tracks scroll, failed runs get a red dot.

**Backend:** `GET /me/agents/{id}` and `GET /me/agents/{id}/runs` (runs grouped from `history_events` by the per-run session prefix now shared with `build_scheduled_turn`; failure flagged via the stored `RUN_FAILED_PREFIX` marker). `agent_chats_only` on `/me/sessions` is deleted along with its only consumer.

**Bug fixes:**
- **In-place tab navigation showed stale content** — clicking a different item of the same kind changed the tab title but not the body, because the body was keyed by tab id only and the reused component kept its state. Now keyed by refId too.
- **`agentIdFromRef` misparsed stored session ids** — every `agent-*` refId was treated as a named agent, giving session tabs a Config side pointing at a nonexistent agent. Now matches the full uuid shape only.
- **`/agents?resume=` deep link was dead** — the old standalone agents page never renders under the workspace shell; resume handling moved into the Workbench URL→tab sync, and the "Resume in chat" button no longer shows for scheduled-run sessions.

## Testing

- New backend tests: runs grouped oldest-first with failure flag + limit; other users' agents 404. Full backend suite: 969 passed; the 3 failures in `test_github_skill_import.py` are sandbox network blocks (`httpx.ConnectError` reaching GitHub), unrelated to this diff.
- Frontend: 234 vitest tests pass, lint clean.
- Browser-verified: collapsed sidebar, runs feed with separators/rail/jump/active-tracking/failed dot, in-place agent switching now swaps content, resume deep link opens the chat with markdown rendered.

https://claude.ai/code/session_01U8KhhTwg6xkSCksPckNTvX


samzliu:
Product screenshots:
- Runs feed at the newest run — reset separators, failed run marked, collapsed prompts, Runs rail with active tracking + failed dot: https://app.joinstash.ai/f/dee39da7-7880-4b69-9c0c-981207d229e9
- After clicking the first rail entry — jumped to the oldest run, rail highlight follows: https://app.joinstash.ai/f/1d08c5bb-b07d-4f67-a7b7-ac4e7a9e49dd
- In-place switch to Stash Agent (previously showed stale content) — chat panel with composer: https://app.joinstash.ai/f/ae1366ee-368d-4bf8-976e-58d123c7914a

PR #915, opened 2026-07-24
Integrations: stop reporting configuration and dead grants as sync failures

samzliu:
Four sync-state bugs found while chasing "why are my X bookmarks stale?". Each surfaced as a red error the user could do nothing about, and none of them retried their way out.

### Slack with no channels reads as a failure
An empty channel allowlist raised `RuntimeError`, so the source showed **failed** with the generic *"Sync hit an unexpected error — it will retry automatically."* Both halves are wrong: nothing is broken, and no retry can help — only the owner picking channels can.

Adds `SourceSetupRequired` → a `needs_setup` status that carries the instruction ("Choose the Slack channels to index"), rendered amber rather than red. The status line also stops showing `last_synced_at` in that state, where a six-week-old date read as *a sync that ran and found nothing* instead of *one that has never run*.

### X bookmarks silently stopped syncing
Bookmarks last updated July 20. X kills a refresh token that idles ~24h, and the daily bookmark gate (#870) left it idle for exactly that: refreshes succeeded every ~2h for days, then the first post-gate refresh got `401` from X's token endpoint and every daily attempt since has failed.

- New `keep_tokens_fresh` beat task (30-min tick) keeps X grants rotating on the cadence X demonstrably tolerates.
- A dead grant now sets a persistent "reconnect X" warning instead of throwing — the old behaviour also blanked that run's posts/replies sync and had its `failed` status erased by the next sync 30 minutes later, so it was invisible in the UI.

A grant already dead still needs one manual reconnect; this prevents the next one.

### GitHub PR discovery retried unreadable PRs forever
`fetch_pull_request` handled 404 but let 403 raise. Sessions referencing PRs in an org the token can't read (`heavi-inc/heavi`) therefore never got their "checked" row, so the 5-minute reconciler re-enqueued the same sessions indefinitely — hundreds of errored tasks a day. Now recorded like a 404; rate-limit 403s still raise so they retry.

### Source browser hid whole folders behind "Load more"
Entries loaded 200 at a time, but folder rows are synthesized from the loaded entries — so an X source with 200+ bookmarks (which sort first) showed **no Posts, Replies, or Articles folders at all** until you clicked "Load all". Now loads every page up front, rendering progressively.

## Testing

- 991 backend tests pass (`pytest`), 237 frontend tests pass (`vitest`), ruff + eslint clean.
- New coverage: Slack-without-channels parks as `needs_setup` with no `last_synced_at`; dead X grant warns while posts keep syncing; keep-warm refreshes live grants and survives dead ones; 403 returns None while rate-limit 403 raises.
- The existing `test_slack_sync_without_channels_records_sync_error` encoded the old "this is a failure" intent and was rewritten in place rather than duplicated.

https://claude.ai/code/session_01NdacdeuYJ4PrPRRK84kS5E

### DIFF
--- a/backend/celery_app.py
+++ b/backend/celery_app.py
@@ -36,6 +36,7 @@ celery = Celery(
         "backend.tasks.sources",
         "backend.tasks.agent_schedules",
         "backend.integrations.google.exporters.slides",
+        "backend.integrations.x_saves.tasks",
         "backend.exports.pdf",
         "backend.exports.pptx",
     ],
@@ -117,6 +118,13 @@ celery.conf.update(
         "sources-reconcile-github-sync-all": {
             "task": "backend.tasks.sources.reconcile_github_sync_all",
             "schedule": 3600.0,
+        },
+        "x-keep-tokens-fresh": {
+            "task": "backend.integrations.x_saves.keep_tokens_fresh",
+            # X kills grants whose rotating refresh token idles for ~a day; a
+            # 30-min tick refreshes each access token as it expires (~2h),
+            # keeping the refresh token exercised (see x_saves/tasks.py).
+            "schedule": 1800.0,
         },
         "cli-auth-cleanup-expired": {
             "task": "backend.tasks.cli_auth.cleanup_expired_sessions",
--- a/backend/integrations/slack/indexer.py
+++ b/backend/integrations/slack/indexer.py
@@ -157,9 +157,11 @@ async def index_slack(source: dict) -> str | None:
     allowed_channel_ids = set(source_service.slack_allowed_channel_ids(source))
     await source_service.purge_disallowed_copied_documents(source)
     if not allowed_channel_ids:
-        # Fail loudly so the sync records a sync_error instead of reporting a
-        # successful sync that ingested nothing.
-        raise RuntimeError("no allowed channels configured")
+        # Nothing is wrong and no retry can help — Slack indexes only the
+        # channels the owner picks, and none are picked yet.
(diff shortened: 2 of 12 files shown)


samzliu:
x

samzliu:
## Verified in the running app

Screenshots captured locally against the real stack (uvicorn + seeded Slack/X sources, driven through `_sync_source`, not hand-set DB rows). They're attached in chat for drag-drop into this thread — `gh` has no path to GitHub's image CDN, and repo policy forbids committing screenshot assets.

**Slack, no channels — before:** red `failed` pill · `synced 6/11/2026 · 0 items` · red *"Sync hit an unexpected error — it will retry automatically."* Every part of that line is wrong: nothing errored, the June date is from the last sync under the old config, and the retry cannot help.

**Slack, no channels — after:** amber `needs setup` pill · `not syncing yet · 0 items` · amber *"Choose the Slack channels to index — nothing syncs until you do."*

**X source browse — after:** all four folders (Bookmarks, Articles, Replies, Posts) visible at once on a 244-item source where 220 are bookmarks. Pre-fix, the 200-row page was entirely bookmarks, so Posts/Replies/Articles did not exist in the UI until you clicked "Load all" — those buttons are now gone.

End-to-end check on the state this PR is really about, run through the actual sync task:

```
before: sync_status='failed'      sync_error='Sync hit an unexpected error — it will retry automatically.'
sync result: {'status': 'needs_setup'}
after:  sync_status='needs_setup' sync_error='Choose the Slack channels to index — nothing syncs until you do.'
```

PR #921, opened 2026-07-27
Bookmarks as a mini program: enrichment, filters, and library skills

samzliu:
## Summary

Makes Bookmarks a first-class app rather than a spreadsheet: rows are summarised and topic-tagged by an LLM in the background, and `/apps/bookmarks` renders them as cards with filters, a detail pane and bulk editing. The generalisation underneath is a **mini program** — a manifest over an ordinary table — so a second one (CRM) is a config file rather than a project.

Also adds four library skills, and fixes several correctness bugs found on the way, two of which would have bitten in production.

## Changes

**Mini programs and enrichment**
- `mini_program_service.py` — a manifest naming a table's schema, enrichment config, seeded views, and which columns carry a row's title/body/labels. Bookmarks is the first; `clip_service` no longer carries its own schema copy.
- `tasks/enrichment.py` — a beat sweep shaped like the embedding reconciler: stale flag, claimed batch, content hash so a re-run import never re-pays. Failures record a reason and leave the queue rather than wedging later sweeps.
- Topic labels feed back into the multiselect column's options, which are simultaneously the model's vocabulary and the UI's filter chips. Bootstrapped from a sample of titles, because labelling row-by-row from an empty vocabulary drifts into synonyms — 5 bookmarks produced 18 near-duplicate topics before this.

**App shell** (`/apps`, `/apps/[slug]`)
- Card grid, detail pane, view switcher, per-row and bulk topic editing with vocabulary autocomplete.
- Server-side paging, search and filtering. Doing them client-side over a loaded page meant a large library silently searched its first 100 items.
- Rail entry relabelled **Views**, moved above the divider. No workbench or explorer changes.

**Raindrop-style filters** — duplicates / broken / untagged as counted chips over the same list, not separate screens.
- Duplicates match on a normalized URL (protocol, `www.`, trailing slash, tracking params); the oldest save is kept as the original.
- The link checker only trusts 404/410 — a 403 or timeout means "couldn't tell", since consent walls return those to non-browser clients constantly.

**Four library skills**, moved out of Python string constants into `backend/skills/<name>/SKILL.md` so a skill can be reviewed as a diff and edited without touching code. Adding a skill is now adding a directory.
- `brief` — what you saved in a period, led by the thread through it rather than an inventory.
- `resurface` — old saves with a specific *why now* from your actual work; refuses anything it can't justify, since the feed already resurfaces at random.
- `overview` — four charts with the reading written next to each.
- `cleanup` — duplicates, dead links, untagged; **proposes, never disposes**.

**`docs/skills.md`** covers the distinction that keeps getting conflated: seeding is what makes a skill *usable* (`list_skills` reads the user's own scope, not the published-skills table), publishing is what makes it *findable*.

## Bugs fixed along the way

- **`0169` used psycopg `%s` placeholders**, but migrations run over asyncpg, which binds `$1`. It passed everywhere because its loop only executes for users who *already had* a Bookmarks table — no test or dev DB did. **In production it would have crashed the backend at boot**, on the migration meant to carry existing users forward. A guard test now scans every migration for it.
- **`query_table` resolved tables by name** and took the first match ordered by `updated_at DESC`. Two tables called "Notes" meant an agent silently read a different one week to week, with no error. Tables now have the per-folder uniqueness pages and folders have always had (`0172`, renaming existing collisions rather than failing), and `query_table` refuses an ambiguous name with the candidates.
- **`TableResponse` had no `views` field**, so FastAPI silently dropped every saved view on load. Pre-existing — the normal table UI couldn't show saved views on first load either.
- **The enrichment sweep double-paid.** `_claim_batch` selected rows without claiming them, and a batch outlasted the beat interval, so overlapping sweeps re-enriched the same rows. Now an atomic `UPDATE … RETURNING` with `SKIP LOCKED`.
- **`0170`** drops `idx_personal_table_unique`, dead since the workspaces removal rewrote its predicate to one no row can satisfy.

## Test plan

- [x] Existing tests pass — **1198 passed**. The 4 failures in `test_activity.py` are pre-existing; confirmed by stashing this branch and running against a fresh database.
- [x] New tests added — 46 across `test_enrichment.py`, `test_skill_seeds.py`, `test_table_name_uniqueness.py`, `test_migration_placeholders.py`. They encode the properties actually regressed while building: concurrent table creation, poison rows, unchanged rows never re-sent, overlapping sweeps, duplicates keeping the original, search spanning the whole table, bulk edits unable to reach another user's rows.
- [x] Frontend — 243 vitest pass, lint and build clean.
- [x] Migration `0172` tested against real collisions: three tables named "Collide" plus a pre-existing "Collide (2)" resolved to four unique names, with the loop catching the secondary collision. Nothing deleted or merged.
- [x] The `%s` guard verified both ways — fails when the bug is reintroduced, passes when fixed.
- [x] Tested manually against a live stack with real model output (`claude-haiku-4-5`): 7 bookmarks enriched, 0 errors, ~4s/row. Browser QA covered gallery → cards → topic filter (5→2) → duplicates filter → bulk bar → tag edit round-tripping to a filter chip → list view → mobile, with zero console errors.
- [x] 100-tab import measured: request returns in **0.38s**, all 100 processed out-of-band. One request, not 100.

## Checklist

- [x] No hardcoded secrets or credentials
- [x] Documentation updated if behaviour changed — `docs/skills.md`
- [ ] CHANGELOG.md updated (for user-facing changes)

## Not in this PR

- **Publishing.** None of the seeded skills — including `slides`, which predates this work — has a row in the `skills` table, so none appear in Discover. Publishing needs an owning account (`skills.owner_user_id`), which is a product decision about whether a "Stash" system account exists.
- **Seeding never updates an existing copy.** Deliberate (never clobber a user's edits) but it means skill content freezes per user at first seed. The fix is seeding *by installing from a published skill*; it needs its own design pass because a fresh or self-hosted instance would then seed nothing.
- **Quota gating on ingest.** `routers/clips.py` still has no per-user limits beyond per-request size caps, with `MAX_IMPORT_URLS = 100_000`. Worth having before the extension ships.
- **`table_service.list_tables`** still has a dead `owner_user_id IS NULL` branch — a workspace-removal leftover, harmless.


samzliu:
Implemented the review fixes and completed a rendered UI pass.

Fixes pushed in `1490de5e`, `c22e391a`, and `5309f151`:
- renumbered the PR migrations after current `main` and made existing-Bookmarks adoption deterministic when an owner has duplicates
- scoped enrichment page/file reads to the table owner
- blocked private/link-local link-check targets and revalidated every redirect
- prevented slow enrichment responses from overwriting concurrent user edits
- serialized concurrent topic-vocabulary updates
- return 409 for rename/move table-name collisions
- apply saved-view filters and sort in the app, with stale-request protection
- made Recent the real default view, formatted dates, preserved loaded rows during enrichment polling
- fixed the mobile detail panel and made touch selection discoverable
- stopped the missing-app state from issuing an unhandled rows request before setup

QA screenshots (kept out of the repo):
- [Desktop Bookmarks view](https://app.joinstash.ai/f/cd65567b-de78-468e-99ea-11d9fb1fc34b)
- [Mobile bookmark detail](https://app.joinstash.ai/f/87477c93-ff6a-4584-ab36-c77dd568ddc8)

Validation:
- backend: 1,213 passed on a fresh database
- frontend: 244 Vitest tests passed
- frontend lint and production build passed
- Alembic: single `0173` head
- live desktop/tablet/mobile browser QA: zero console errors

samzliu:
Implemented the remaining review follow-ups in `64ead136`.

### What changed

- Moved Bookmarks paging fully server-side: saved-view filters, search, topic filters, derived filters, totals, sorting, limit/offset, duplicate ranking, and facets now run in Postgres. Normal page/search requests skip duplicate normalization.
- Added a first-use Bookmarks empty state with a clear extension CTA; filter-empty results remain distinct.
- Removed automatic Skill distribution:
  - no signup seeding
  - no hourly Celery backfill
  - no Slides-specific built-in prompt path
  - no standalone Slides backfill script
- Kept only `brief`, `cleanup`, `overview`, `resurface`, and `slides` as documentation drafts under `docs/skills/`.
- Existing user-installed Skill copies are intentionally untouched. Publishing these drafts through the existing service-account/public-Skill workflow is outside this PR.

### Product screenshots

- [Desktop onboarding](https://app.joinstash.ai/f/6ddf6459-ac73-4d60-aba9-04422d7a9948)
- [Mobile onboarding](https://app.joinstash.ai/f/a674bc06-95b7-408c-9a73-2ddc0d47c985)

### Validation

- Backend: **1,205 passed**
- Frontend: **244 passed**
- Backend Ruff checks: passed
- Frontend production build: passed
- Frontend lint: 0 errors (3 unrelated pre-existing warnings)
- Browser QA: empty account, desktop/mobile empty state, CTA navigation, and no browser errors
- 10,000-row local exercise: server returned only the requested page; search, duplicate filtering, and facets stayed database-side

henry-dowling (approved):
Yep go for it! Blast radius much smaller than the diff would make it seem

PR #923, opened 2026-07-27
fix: link Gong documents to original calls

cheese-cakee:
## Summary

- add canonical Gong call URLs to connected-source documents
- return the link only after existing ownership and workspace-allowlist checks pass
- cover both URL construction and the authorized document-read path

## Why

Gong calls were indexed with their stable call IDs, but Stash returned no provider URL for them. Users could read and search a call in Stash without a direct way to open the original in Gong. Gong documents the call page format as `https://app.gong.io/call?id={id}`, so the existing stored path is sufficient and no schema change is needed.

## Impact

Gong source documents now include a canonical link back to the original call. Other source providers are unchanged.

## Validation

- `python -m pytest backend/tests/test_sources.py::test_gong_visibility_requires_account_allowlist backend/tests/test_sources.py::test_source_document_url_builds_provider_deep_links -q --no-cov` (2 passed)
- `python -m pytest backend/tests/test_sources.py -q --no-cov` (104 passed)
- `ruff check backend/ cli/`
- independent review: no findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ everything before it is captured in git history (`git log`), not here.
 
 ## Unreleased
 
+- Gong call documents now link back to the original call in Gong.
 - `stash vfs stat` once again shows the source-sharing command for connected
   source roots, including roots that do not have an app URL.
 - OAuth reconnects now require a stable provider account identity. Slack,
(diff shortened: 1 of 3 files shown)


ColemanDuPlessie (approved):
This looks good, thank you for resolving this! One tiny note: When you edited the comment saying slack, granola, and gong needed additional data to reflect that you added the link for gong, you dropped the TODO, so cmd+f for "TODO" wouldn't pick it up anymore. This is fairly trivial, so I've just made a quick 1-line commit that edits that comment for you and approved your PR with the edited comment.

PR #925, opened 2026-07-28
fix: resolve .mcp.json sync path for local exec mode

ColemanDuPlessie:
## Abbreviated Summary

When running locally, curator agents (and other schedule agents) would crash due to not having permission to write to home/sprite). This PR adds a check to switch to ~/.stash-dev-sprite instead if running locally.

## Summary (AI-generated below this line)

Every scheduled agent run in local dev mode (`AGENT_EXEC_MODE=local`, the default) crashed with `SpriteError('write_file failed for /home/sprite/work/.mcp.json')` before the agent even started.

`sync_sprite_config` (added in #884) hardcodes the Sprites-VM workdir path when writing the MCP registry as `.mcp.json`. The local exec layer translates only the **cwd** (`/home/sprite/work` → `~/.stash-dev-sprite/work`), not **argv**, so the write helper literally ran `os.makedirs('/home/sprite', ...)` on the developer's machine — `/home` is a root-owned read-only autofs mount on macOS, so the child exits 1 and the run dies.

## Changes

- New `sprite_service.write_workdir_file(sprite, rel_path, contents)`: writes a workdir-relative file, resolving the path with the existing mode-aware `_box_path` (Sprites → `/home/sprite/work`, local → the simulated workdir).
- `sync_sprite_config` uses it instead of hardcoding `SPRITE_WORKDIR`.
- Regression test asserting the resolved path in both exec modes.
- Sprites-mode behavior is unchanged (`/home/sprite/work/.mcp.json`, same as before).

The OAuth credential-file writes also hardcode `/home/sprite` paths, but local mode never injects credential files (`agent_auth` returns bare `RunAuth(harness=CLAUDE)`), so `sync_sprite_config` was the only local-mode caller affected.

## Verification

- `pytest backend/tests/test_sprite_agent_mapping.py`: passed, including the new `test_write_workdir_file_resolves_per_exec_mode`
- `pytest backend/tests/test_mcp_servers.py`: 11 passed; `test_registered_servers_reach_the_sprite_mcp_config` fails identically on clean `main` in this environment (pre-existing, unrelated)
- `ruff check` / `ruff format --check` on changed files: passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ everything before it is captured in git history (`git log`), not here.
 
 ## Unreleased
 
+- Scheduled agent runs no longer crash in local dev mode: the MCP registry's
+  `.mcp.json` is now written to the local simulated workdir instead of the
+  literal `/home/sprite/work` path, which is unwritable on dev machines.
 - Gong call documents now link back to the original call in Gong.
 - `stash vfs stat` once again shows the source-sharing command for connected
   source roots, including roots that do not have an app URL.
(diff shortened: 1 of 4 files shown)


henry-dowling (approved):
LGTM

PR #930, opened 2026-07-28
refactor: reads span every scope you own (scope as a place, not a mode)

henry-dowling:
## Why

`X-Stash-Scope` is a **mode**, not a place. One key in `~/.stash/config.json` silently re-roots what every read returns, so the same `GET` answers differently depending on a header someone set days ago on another machine. Two concrete consequences:

- A session streamed into a workspace becomes unfindable from a personal scope, even though you own both.
- The CLI cannot search the team KB while streaming personally — the single `scope` key drives reads *and* writes (`cli/client.py:88`), so they can never diverge.

The fix is to stop treating scope as ambient state. **You own one scope per membership — yourself, plus each workspace — and those are places.** Reads span all of them and tag each row with the scope it came from; the client renders those as roots (`/personal`, `/workspace/<domain>`). Only writes still choose a single scope, because a new object has to land somewhere.

## Scope of this PR — please read

This is **increment 1 of 4**. The full read-side conversion is ~1,700 `owner_user_id` sites across 11 routers and 8 service modules; that is not one reviewable PR, and a half-converted read path is a broken app. This PR lands the primitive plus one reference conversion per read shape, so the pattern is settled before the bulk lands.

Converted here (3 of 43 read routes):

| Route | Shape | Change |
|---|---|---|
| `GET /me/sessions` | paginated span | Drops the ambient narrowing; already spanned via `accessible_scope_ids_sql` |
| `GET /me/sessions/{session_id}` | by-id | Resolves which owned scope holds the id instead of being told |
| `GET /me/trash` | unpaginated fan-out | Per-scope listing, merged and sorted, rows tagged |

The three shapes exist because **pagination decides the technique**: paginated lists must span in SQL (merging pages of separate queries is not one list), unpaginated listings can fan out and sort in Python, and by-id reads resolve the owner. Every follow-up route falls into one of these.

Remaining increments, for sequencing — not started:

2. The other 40 read routes: `files_tree` (24), `tables` (10), `files` (3), `memory` (4), `sources` (4), `skills` (3), `user_knowledge` (2). Tree endpoints (`/tree`, `/sidebar`, `/overview`) fan out per scope and nest under two roots — a two-root tree is literally two trees.
3. The 64 mutating routes: resolve the owner from the object id (`resolve_owner_user_id` already exists) rather than from the header, so editing a workspace page doesn't require pre-selecting the workspace. Creation keeps `get_scope`.
4. Clients: `/personal` + `/workspace/<domain>` roots in the VFS (`stashvfs/model.py` roots are already synthesized, so they slot in), frontend switcher becomes navigation rather than a mode, and the CLI's read/write coupling disappears. Documented paths (`CLAUDE.md`, `stash prompts agent-guidance`, README) change in one shot — no aliases.

## What is deliberately unchanged

- **Ownership and permissions.** `owner_user_id` stays the security boundary. Spanning reuses `workspace_member_condition`, so it cannot widen access by one row — an unverified on-domain signup still gets nothing, which is covered by a test.
- **Writes.** `get_scope` and the plugin's streaming path are untouched, so the workspace stays the durable owner of what a member streams into it.
- **Shares.** Excluded from the owner set on purpose: shared-in content is somebody else's scope, surfaced per object by `readable_content_condition`, not a root of your own.

## Testing

`backend/tests/test_read_scopes.py` — 11 new tests, all passing:

- the owner set: no workspace, member ordering, **unverified on-domain signup gets nothing**, non-member gets nothing
- spanning reads: trash spans and labels rows, trash still hides other scopes, session detail resolves the scope, session detail 404s for scopes you don't own, session list spans with no header, the header no longer narrows, an explicit `owner_user_id` filter still narrows

Regression run across the 12 suites that touch these routes or `X-Stash-Scope`: **172 passed, 1 failed**. The failure is `test_workspaces.py::test_admin_creates_workspace_with_working_bootstrap_key` (`column "auth0_sub" does not exist`) and is **pre-existing** — it fails identically on a clean tree, verified by stashing this branch's changes. `ruff check` and `ruff format --check` clean.

No GUI changes in this increment, so no screenshots — the frontend still renders the switcher as it does today. Increment 4 is where screenshots will matter.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -46,9 +46,19 @@ S3-backed binary, vs. an in-app-editable page, vs. a folder).
   integrations keyring. The cloud agent's sprite gets the registry written
   into its `.mcp.json` every turn.
 - A workspace is an org-owned, login-less scope (`workspaces.domain`,
-  membership derived from verified email domain). Content routes select the
-  scope via the `X-Stash-Scope` header — including the plugin's session,
-  event, transcript, and artifact writes — and non-members are hard-403'd.
+  membership derived from verified email domain). Non-members are hard-403'd.
+- **Scope is a place, not a mode.** A caller owns one scope per membership —
+  themselves, plus each workspace — and those are the roots a client renders
+  (/personal, /workspace/&lt;domain&gt;). Reads span all of them
+  (`auth.get_read_scopes` → `permission_service.read_scope_ids`) and tag each row
+  with its `owner_user_id`, so what a GET returns never depends on ambient state.
+  Writes still resolve to exactly one scope (`auth.get_scope`, selected by the
+  `X-Stash-Scope` header) because a new object has to land somewhere: that is the
+  plugin's session, event, transcript, and artifact destination.
+  - Three read shapes, by pagination: paginated lists span in SQL
+    (`accessible_scope_ids_sql`, since merging separate pages is not one list);
+    unpaginated listings fan out per scope and sort in Python; by-id reads
+    resolve which owned scope holds the id rather than being told.
 
 Object-level privacy tables and page-link graph tables are intentionally not part
 of the current architecture. Privacy is mediated by Skills.
--- a/backend/auth.py
+++ b/backend/auth.py
@@ -264,6 +264,23 @@ async def get_scope(
     return scope_user_id
 
 
+async def get_read_scopes(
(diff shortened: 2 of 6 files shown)


henry-dowling:
I'm gonna close this since we're reworking workspaces anyway

PR #939, opened 2026-07-29
Skills as a launcher, plus the two dead surfaces it uncovered

samzliu:
A skill was only ever a folder you could read. Running one meant knowing it existed, knowing its name, and typing that name at an agent — so the library skills shipped in #921 had no way to reach a user at all. Fixing that turned over two adjacent rocks, both of which are in here.

> **Heads up:** this branch also carries two commits that aren't part of the skills work — `feat: make bookmarks a table-first app` and `feat: add bookmark layouts and sorting`. They landed on the branch mid-review. Everything below describes the rest.

## 1. Skills you can run

**Publishing.** The curator importer took a whole repo, and this repo holds `SKILL.md` files that aren't library skills (`backend/static`, `www/public/smb`). It now accepts a `/tree/<branch>/<dir>` suffix, so this publishes five skills and nothing else:

```
POST /api/v1/admin/discover-skills/import
{"repo_url": "https://github.com/Fergana-Labs/stash/tree/main/docs/skills"}
```

**This has to be run after merge.** Publishing is deliberately not a signup seed, startup hook, or periodic backfill. Re-running tracks upstream by `source_github_url`, so slugs and view counts survive an edit.

**Adding is always explicit.** An agent resolves skills from its own scope, never from the Discover catalog, so a published skill has to be added before anything can run it. `POST /me/skills/install` does that, idempotently by name. Running never installs as a side effect: the launcher only opens on skills already in your Skills, Discover cards keep Open/Save, and the strip above a mini program offers Add for what you don't hold.

**The launcher composes nothing.** It shows the skill's own `description` and `when_to_use` and sends `Use the <name> skill.` ahead of your request. That one templated line exists because the agent otherwise picks skills by matching your message against every `when_to_use` — without naming it, Run means "probably this skill". The skill format gained no fields for any of this.

**The Skills page was dead code.** `/skills` is a workbench section and wasn't in `rendersRouteContent`, so the route rendered "Open something from the explorer" and `page.tsx` never mounted — the exact failure `workspace-shell.test.tsx` was written to catch, which it didn't, because the test asserted the broken behaviour. Registering the route is what makes the launcher reachable; it also surfaces the tabs, pins and Discover grid for the first time, so that's more new surface than the launcher alone.

## 2. "Shared with me" had no UI at all

Same dead-route bug, worse consequence. Files' shared scope lives on `files/page.tsx`, `/files` isn't in `rendersRouteContent` either, and `SharedWithMeFiles` is imported by exactly one component — the dead one. The Files explorer, which does render, had no shared surface. So `GET /shares/with-me` worked and nothing in the app could reach it.

Sessions was the only section that solved this, by merging shared folders into its explorer root. Files and Skills now do the same: a **Shared with me** node at the root of each tree, tagged with the owner because names collide across people.

The node is an index, not a folder — sentinel id, no tab view, hides New/Upload/Import while you're inside it, takes no drop, isn't draggable. Items under it are `readOnly`, so Rename/Delete stay hidden. Skills' old "Shared with you" tab is gone, replaced by the node: a shared skill was never in your VFS, so your agent can't run it until you copy it in, exactly like a shared folder.

## 3. Clips could be renamed and deleted

Memory has been protected since it got `is_memory`. Clips had nothing — verified against a running stack, rename returned `200` and delete returned `204`. Neither fails cleanly: `clip_service` finds Clips by name at the scope root, so a renamed one is invisible to it and the next save silently starts a fresh empty folder; a deleted one takes every clip with it.

Both are the same kind of thing — resolved by identity, written into by code the user never sees — so they get one flag (`is_protected`) and one guard rather than a per-folder special case. The guard is in the service, not the router, because the UI isn't the only caller: the CLI, the VFS, and the agent's tools all reach these functions. Migration `0174` backfills existing rows the way `clip_service` identifies them, so a nested folder you called "Clips" stays yours.

## Screenshots

**https://app.joinstash.ai/skills/pr-939-skills-launcher-shared-with-me-protected-folders-sf532g**

On Stash rather than inline — anonymous fetch of a file in a non-discoverable skill 404s, so `` tags here would render broken.

1. `s1-skills-yours.png` — the Skills page, rendering for the first time. Run on a skill you hold.
2. `s2-discover-no-run.png` — Discover has no Run; you add first.
3. `s3-launcher.png` — the launcher: description, when_to_use, your request. Nothing authored for it.
4. `s4-bookmarks-strip.png` — the Bookmarks strip: `+ brief` (not held) beside `resurface` (held).
5. `s5-protected-clips.png` — Clips' context menu: no Rename, no Delete, `draggable=false`.
6. `s6-shared-node.png` — Files → Shared with me → `Q3 plan (Henry)`.

## Verification

- 1232 backend tests, 270 frontend tests, ruff + eslint clean.
- Verified end to end on a local stack: subpath import published 5 skills (not the 7 an unscoped import grabs); Add put `resurface` in a fresh account and a second Add returned the same folder; the run reached the agent as `Use the resurface skill.` + the request; a folder shared from a second account appeared under Files → Shared with me while the shared *skill* appeared under Skills and not Files; `PATCH`/`DELETE` on Clips now both return `400` with `the Clips folder can't be renamed, moved, or deleted`.
- Writing the "you can't move things into Shared with me" test found that you could — the row ignored the drop but the list container behind it caught it. Fixed in `7fbe3be`.
- Pre-existing and untouched: `test_github_skill_import`'s three network-dependent tests fail in a sandbox without outbound network; a TS error in `page.test.tsx`; `seed_dev` raises `KeyError: 'sync_warning'`.

https://claude.ai/code/session_01Th6i5wXAcrWjZCHVyuh6jh


samzliu:
Integrated the bookmark-manager UI cleanup in `2cededb1`.

What changed:
- Sidebar now says **Apps**, and the Apps gallery no longer shows the redundant marketing tagline.
- Bookmarks always opens as a compact table with Title, URL, Type, Saved, Site, Clip, Summary, Topics, and Status.
- Removed the duplicate in-app search and Recent / By topic / card-view chrome.
- Clicking a title opens an editable right drawer for bookmark fields and generated topics.
- Shift-click now selects a contiguous row range for bulk actions.
- Preserved this PR’s **Put these to work** skills banner above the new table.

QA screenshots:
- [Table-first Bookmarks view](https://app.joinstash.ai/f/b91aff9b-3a59-4af2-8aeb-41b9e81687ed)
- [Editable right drawer](https://app.joinstash.ai/f/509db2e4-d1c8-47e5-a868-2c62dd19a8b9)

Verification on the combined PR branch:
- `npm test`: 59 files, 262 tests passed
- `npm run lint`: 0 errors (one unrelated existing warning in `src/app/search/page.tsx`)
- `npm run build`: passed
- Authenticated browser QA: topic edit persisted, title edit persisted, shift-range selected 3 rows, no console errors

samzliu:
Follow-up bookmark UI revision in `31be376c`:

- Restored **Cards** behind a Table/Cards toggle; Table remains the default.
- Added ascending/descending sorting for every column. Sorting runs server-side across the full filtered result set, not only the loaded page.
- Changed the desktop details drawer into a dedicated 420px grid column, so it no longer covers the table.
- Preserved generated-topic editing, shift-range selection, and this PR’s skills banner.

Verification:
- 59 frontend test files, 262 tests passed
- ESLint: 0 errors
- Production build passed
- Ruff check/format passed
- 2 focused backend sorting tests passed against an isolated Postgres database

The screenshots in the previous comment show the superseded table-only layout and should be ignored. Updated screenshots are pending the local browser QA tool upgrade.

samzliu:
## Extension popup changes (v0.5.0)

Saving had no working confirmation of any kind. Every OS notification was dead (relative `iconUrl` resolving against `dist/` — see commit body), the popup never rendered `lastClip`, and save-all's success path did nothing. Fixed, plus the links-vs-contents wording.

**Save this tab** — button goes to `Saving this tab…` (grey, no tick), then:
> ✓ Saved "The Bitter Lesson"
> View in Stash

**Save all open tabs** — button face names the scope up front: `Save all 23 open tabs`, counted through the same http(s) filter the save itself uses. Click acknowledges immediately, before the request returns:
> Saving 23 tabs…
> You can close this popup — it keeps going without you.

then on completion:
> ✓ 23 links saved
> Downloading page contents now. 4 were already in your Stash.

**All duplicates** (the case that produced "0 tabs saved to Stash"):
> ✓ Nothing new to save
> All 6 open tabs are already in your Stash.

No batch is created, so no phantom progress row.

**Progress row** now reads `Page contents: 5 of 23 downloaded, 18 pending` — previously "5 saved", which collided with the links count — and retires once its import has finished and announced itself instead of parking indefinitely.

Verified by running the built `popup.js` in a browser against a stubbed `chrome` API and clicking through each state; screenshots are in the working session. (No agent path for attaching images to a PR — @samzliu these need dragging in manually if you want them on the description.)

henry-dowling (approved):
Go for it!

PR #961, opened 2026-08-03
Make scheduled-agent skips legible: record last_run_outcome per tick

henry-dowling:
This is a small change to our logging- we explicitly log the performance of our nightly curator agent so we can continuously monitor the health of our system.

# Slop
## Why

The nightly curator's beat task consumes the cron tick up front (`mark_run` stamps `last_run_at`), after which a run either executes, fails, or is skipped by one of three designed gates: credit allowance, missing credential, or the no-changes cost gate. Only failures left a trace (`last_run_error`), so from the database a healthy skip was indistinguishable from a run that died mid-flight.

This has now caused two false alarms for the same customer: the July "curator silently stopped" incident, and Aug 2, when Heavi's curator stamped `last_run_at` and then went dark — hours of diagnosis later, it turned out to be the no-changes gate correctly skipping an idle weekend account.

Worse, the stale-curator watchdog keyed off `last_run_error IS NOT NULL` — precisely the field a mid-flight death never sets — so the one failure shape with no other trace was also the one the watchdog couldn't see.

## What

- **Migration 0174**: `agents.last_run_outcome text`.
- `mark_run` stamps `'started'`; the tick must then resolve it:
  - `mark_run_succeeded` → `'ran'`
  - `mark_run_failed` → `'failed'` (alongside the existing error text)
  - new `mark_run_skipped` → `'skipped_credits' | 'skipped_no_credential' | 'skipped_no_changes'`
  - a row stuck at `'started'` now means exactly one thing: the run died without resolving.
- All three skip gates in `_run_due` record their reason; the beat path and the manual-run path (`_run_curator_now`) both record success.
- `alert_stale_curators` now also pages on `last_run_outcome = 'started'` (stale watermark + pending changes + a tick that began and never resolved). Designed skips stay quiet, as before.

No metering changes: skips still consume the month credit exactly as they did.

## Tests

- New: silent-death shape alerts (`'started'`, no error); designed skip stays quiet; `_run_due` records `skipped_no_changes` (with `run_scheduled` rigged to fail loudly if the gate leaks).
- Extended: the run-failure test asserts `outcome='failed'`.
- Full backend suite: **1223 passed**, coverage 80.32%, ruff clean.

### DIFF
--- a/backend/migrations/versions/0174_agent_last_run_outcome.py
+++ b/backend/migrations/versions/0174_agent_last_run_outcome.py
@@ -1,0 +1,30 @@
+"""Record how each scheduled-agent tick resolved.
+
+mark_run consumes the cron tick up front (last_run_at), after which a run
+either executes, fails, or is skipped by one of three designed gates (credit
+allowance, missing credential, no changes since the watermark). Only failures
+left a trace (last_run_error), so from the outside a healthy skip was
+indistinguishable from a run that died mid-flight — the confusion behind the
+July 2026 "curator silently stopped" incident, and again on Aug 2 when a
+correct no-changes skip for an enterprise customer read as a silent death.
+
+last_run_outcome makes every tick legible: mark_run stamps 'started', which
+then resolves to 'ran', 'failed', or 'skipped_<reason>'. A row still at
+'started' means the run died without resolving — the stale-curator watchdog
+pages on that. NULL means no tick since this column shipped.
+"""
+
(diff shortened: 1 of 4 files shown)


henry-dowling:
Superseded by #996, which re-implements this outcome state machine on current main, migration 0184, and the current heavy-queue scheduler boundary. Closing this stale branch in favor of the replacement.

PR #965, opened 2026-08-03
Session titles live on sessions; search names sessions by title; titles never contain quotes

ColemanDuPlessie:
Ports three changes made in our benchmarking repo, centered around reorganizing how we deal with session titles:

- `session_titles` table remove, now is a part of `sessions`
- `stash search` CLI command now gives session titles, rather than IDs. The underlying endpoint still returns both, since web search still gets sessions by IDs.
- Quotes are stripped from session titles, to prevent the ugly case where we force agents to use triple-nested quotes to make the query they want (e.g. stash vfs "grep ... 'my title that "has quotes" ...' ...")

AI-generated longer summary below...

## 1. `session_titles` folded into `sessions`

The side table was 1:1 with sessions and FK'd to it — every reader had to LEFT JOIN it, every writer upserted against a table that couldn't exist without its sessions row. Migration 0179 adds `title`, `title_source_hash`, `title_user_set`, `title_updated_at` to `sessions`, carries existing titles forward in one shot, and drops the table. All consumers (title service, generation task, share service, sidebar etag) now read and write the sessions columns; `set_user_title` fails loud if the session row doesn't exist.

## 2. `stash search` names sessions by title

Session search hits now carry `name` — the session's display title in the VFS's spelling (new `session_title_service.titles_for_session_ids` + public `stashvfs.safe_name`) — so a hit can be followed straight into `/sessions/<name>/`. The CLI prints the title and omits the raw session id, which isn't a usable handle there.

**Deliberate deviation from the benchmarking commit:** it removed `ref` from session hits entirely. Here `ref` is kept — the web search page (`frontend/src/app/search/page.tsx`) links sessions via `hit.ref` and skips ref-less hits altogether, so dropping it would have removed sessions from web search. No frontend changes needed.

## 3. Stored titles never contain quotes

Generated, derived, and user-set titles all pass `_strip_quotes` (removes `'`, `"`, and backticks). Stripping happens *after* reply-shape detection, whose patterns need the apostrophes ("I'll …", "you're …") intact. VFS names additionally strip `$` — titles become VFS directory names, which agents reach through two layers of shell parsing in `stash vfs "<script>"`.

## Testing

- Full backend suite green on a fresh test DB: 1267 passed (one deselected: `test_admin_creates_workspace_with_working_bootstrap_key`, which fails identically on main — pre-existing `auth0_sub` env issue).
- CLI suite: 210 passed, 1 pre-existing failure on main too (`test_ls_unknown_source_fails_loudly`, console-rendering assertion).
- Migration 0179 applies cleanly from scratch and on a DB with existing titles; downgrade recreates the table and copies titles back.
- New tests: rename strips quotes end-to-end; `clean_generated_title`/`title_from_text` strip quotes; VFS drops shell-hostile characters from session dir names.

No GUI changes — the web search page's behavior is intentionally unchanged.


henry-dowling (changes requested):
(1) Getting rid of `session_titles` table: love it, great call

(2) `stash search` returns titles of sessions- I have a couple thoughts:
- I believe we unintentionally affect our search rankings here! Look- now we rank based on both title *and* snippet, not just snippet. was this intended? if not, we should keep existing ranking behavior.
- why hide the session ref when searching for sessions? why not include both the ref *and* the title? refs are nice because they're guaranteed to be unique 
- this also introduces inconsistency between our API and CLI- the API returns both, and the CLI returns only the name after this change.
- Edge case to remember- when we have two separate files in the VFS with the same name, we'll automatically disambiguate them with a UUID. Eg
  /sessions/Fix auth--12345678
  /sessions/Fix auth--abcdef12

this is important because it means if you search just using the name of the file you're looking for, it may fail if there were two files created with this name.

(3) I think we're doing this at the wrong place- we're sanitizing titles at import- why not sanitize titles only when we're about to display them in the VFS? This will let us preserve the important invariant that if you rename a session, its name in Stash is the same name that you typed in.  

Basically, the overall takeaway for (3) is- we should persist the ground truth data, not anything that we derive from it. This is so that we don't lose information (that might be important!)

ColemanDuPlessie:
Just updated this, resolving your comment: stash search now returns both title and ID, and titles are sanitized at use, rather than at import.
I also made the CLI commands that take IDs (e.g. stash share) work with IDs or session titles.

I did not, however, allow CLI users to lookup sessions by ID (since currently the system is sessions/[title]/transcript.md): we could add some sort of alias system, but that seems heavy and expensive, so I'm not sure if it's a net win.

ColemanDuPlessie:
Updated again: ID is now still mentioned parenthetically in search results, but is entirely redundant (in the CLI; it's still used to generate unique URLs). Everything that models could do in the CLI with a pre-existing transcript's ID can now also be done with the session title.

henry-dowling:
Closing the loop on this: Confirming that I think we can drop session_id from CLI search results; agents don't really need it for anything

henry-dowling (approved):

PR #967, opened 2026-08-06
embeddings: one provider — OpenAI-compatible or nothing

henry-dowling:
## What

Deletes the `huggingface` and `local` embedding providers and the `EMBEDDING_PROVIDER` auto-detection chain. The embedding service now has exactly one codepath: `OpenAICompatEmbedder` — OpenAI by default, any `/v1/embeddings`-compatible endpoint via `EMBEDDING_API_URL`.

Net: +59 / −326.

## Why

Which company receives customer content for embedding was a function of which API keys happened to exist in the environment (`OPENAI_API_KEY` → OpenAI, else `HF_TOKEN` → Hugging Face, else local). That made the deployed behavior impossible to determine from the code — we spent a chunk of today's SOC 2 policy work establishing empirically (via prod worker logs) that prod runs OpenAI. It's also a fallback chain, which CLAUDE.md bans outright.

Prod has run the OpenAI path (`text-embedding-3-small` @ 384 dims) since the original `embedding_service.py`; the deleted branches were unused there.

## Behavior

- No key → embedding features report unconfigured and stay off (semantic search 503s, indexing marks rows `embed_stale`). Nothing substitutes silently.
- A leftover `EMBEDDING_PROVIDER=local` / `=huggingface` raises at first use with a message explaining the removal — a self-hoster upgrading with the local provider configured must not have their content silently rerouted to an external API.
- `set_embedder` survives as an explicitly-labeled test seam.
- Self-hosters: point `EMBEDDING_API_URL` at any compatible endpoint; `.env.example` and the compose comment updated.

## Prod impact

None. `EMBEDDING_PROVIDER` is not set in prod (provider was chosen by auto-detect finding `OPENAI_API_KEY`), so the new code takes the same path prod already runs. No env changes needed.

## Testing

- `backend/tests/test_embeddings.py` + `test_embedding_security.py`: 5 passed (two tests covering the deleted providers removed)
- `ruff check` + `ruff format --check`: clean
- App import smoke test passes


henry-dowling:
Closing in favor of a docs-only note: the pluggable providers were deliberate (self-hosting feature), so we're keeping them and documenting that prod runs the openai provider instead.

PR #970, opened 2026-08-06
fix: X bookmarks import stranded at a fraction of users' bookmarks — page at 20, re-walk history

henry-dowling:
hotfix to our bookmark import feature. A silent bug in the X API was causing imports to fail. to fix this, we decrease page size (apparently this fixes the bug) and then also persist bookmarks after every page as we walk through the user's bookmarks

## Incident

A user reported publicly (X, Aug 5) that the bookmarks import "choked" partway through their X bookmarks.

## Root cause

X's `GET /2/users/:id/bookmarks` silently drops `meta.next_token` after 1–3 pages at `max_results=100` even when thousands of bookmarks remain — a [known X API bug](https://devcommunity.x.com/t/bookmarks-api-v2-stops-paginating-after-3-pages-no-next-token-returned/257339); the community workaround is smaller pages. Verified live against the affected account with its OAuth token (count-only probe): at 100/page pagination dies at ~110 items; at 20/page X serves all 2,895 across 145 pages and ends with a genuine short final page. The indexer's history walk treated the missing token as "all ingested" and stamped `x_bookmarks_complete`, so it never looked again.

## Fix (`backend/integrations/x_saves/indexer.py`)

- `BOOKMARK_PAGE_SIZE` 100 → **20** — the only size that demonstrably paginates to the end.
- History walk gets its own budget, `MAX_BOOKMARK_WALK_PAGES = 150` (~3,000 bookmarks/daily check; the endpoint allows 180 req/15 min per user, measured). Bigger archives resume next check.
- Cursor is persisted after **every** page, so a mid-walk stop (budget, rate limit, worker redeploy) resumes instead of refetching — and re-billing — walked pages.
- A missing `next_token` on a **full** page is treated as the truncation bug: the cursor stays put and the next daily check retries. Only a **short** final page stamps `x_bookmarks_complete`.

## Migration (`0179_rewalk_x_bookmarks.py`)

One-shot: strips `x_bookmarks_complete` / `x_bookmarks_cursor` / `x_bookmarks_checked_at` from existing `x_saves` sources, so each re-walks from the top on its next sync after deploy. Inserts are idempotent per tweet (`ON CONFLICT DO NOTHING`) — no duplicates. Note: the re-walk bills roughly 3,000–4,000 X API post-reads across the three accounts on day one; that is the point of the fix.

## Tests

Two new tests encode the failure mode: a full page without a `next_token` must not be believed (the 109-of-2,895 scenario, including recovery when X later paginates properly), and the walk must resume from its parked cursor at the small page size without refetching walked pages. Full backend suite: **1,284 passed**, ruff clean.

### DIFF
--- a/backend/integrations/x_saves/indexer.py
+++ b/backend/integrations/x_saves/indexer.py
@@ -59,13 +59,20 @@ MAX_USER_TWEET_PAGES = 5
 MAX_TIMELINE_BACKFILL_PAGES = 25
 # The X API bills per post returned (unlike twitterapi.io, which is cheap),
 # and most checks find no new bookmark — so bookmarks are checked at most
-# once a day, starting with one small probe page. Backlog and history pages
-# use the full size, capped per check. Posts/replies/articles are unaffected:
-# they ride twitterapi.io on the source's normal sync cadence.
+# once a day, starting with one small probe page. Posts/replies/articles are
+# unaffected: they ride twitterapi.io on the source's normal sync cadence.
 BOOKMARK_CHECK_INTERVAL = timedelta(days=1)
 BOOKMARK_PROBE_SIZE = 10
-BOOKMARK_PAGE_SIZE = 100
-MAX_BOOKMARK_PAGES = 5
+# X's bookmarks pagination silently drops next_token after a page or two at
+# max_results=100 even when thousands of bookmarks remain (confirmed on a
+# real account: 100/page ended at ~110 of 2,895; 20/page served all of them).
+# Small pages are the only size known to paginate to the true end.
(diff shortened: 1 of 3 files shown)


henry-dowling:
**Emergency change record (SOC-2 change management):** merged by CTO via documented ruleset bypass under the Operations Security Policy's emergency-change provision ("Emergency changes may be expedited but must undergo a retrospective review and authorization"). Context: production data-completeness incident reported publicly by an affected user; fix verified by full CI (18/18 checks green, 1,284 backend tests) and a live read-only reproduction against the affected account before merge. **Retrospective review requested from the team as of this comment — please review post-hoc and record approval here.**

PR #971, opened 2026-08-06
X grants: pre-expiry token rotation on the heartbeat + refresh forensics

henry-dowling:
This is a small but net good change that was surfaced when I was investigating why Joshua Penman's X refresh token was revoked by X.

We previously had two periodically launched workers on celery: (1) a sync that happened every 30 minutes, which would refresh the expired access token once every 24 hours (2) a "check on token" workflow that would happen once every 30 minutes, and would only refresh an access token if it's within 60 seconds of expiring though, so it therefore missed the token before it expired 97% of the time. However, if it encountered an expired access token, it would attempt to refresh it. So therefore tokens were valid approximately 88% of the time, since they only expire once every 2 hours.

Now, we refresh access tokens before they expire always. This makes our system easier to reason about and makes future debugging easier (case in point: claude got confused and thought that the fact that access tokens were expired when we tried to refresh them, and this was making the X API mad at us. I don't really buy this explanation, since Sam had been using this feature for a long time without any issues, including many refreshes featuring expired tokens, but now we can just not even have that hypothesis in possibility space next time things break).

# Slop

## What this is (and isn't)

**This PR is instrumentation plus a cheap conservative hedge — not a claimed fix for X grant deaths.**

Forensics on all three dead X grants (Alex Jul 29, Sam Aug 4, Joshua Aug 6) ruled out the code-level suspects: the advisory-lock single-flight has existed since Jun 10 (no double-refresh race), the worker had zero unplanned deaths in the past week (no lost-rotation-on-kill), and each dead refresh token was presented exactly once, then refused 401. Best current attribution: Joshua — anomalous diagnostic traffic against his grant (off-infrastructure probe burst, our mistake, policy-fixed); Alex — grant died at its first-ever refresh ~2h after connect, plausibly user-side revocation, unknowable from our side; Sam — one genuinely unexplained refusal in ~650 refreshes over two weeks (~0.15% per-refresh hazard).

So the honest state: at most one confirmed spontaneous death of a healthy, actively-refreshing grant. Rare — but undiagnosable today, because a refusal is a bare 401 and reconstructing a grant's rotation history takes hours of cross-referencing httpx log lines against DB timestamps.

## Changes

1. **Refresh forensics (the actual deliverable).** Every refresh attempt now logs provider, user, `idle_s` (seconds since the last rotation), and outcome (new expiry, or exception type on refusal). The next grant death becomes one log query showing the grant's full rotation history and the exact context of the refusal, instead of an autopsy.

2. **Per-provider `refresh_margin`; X rotates pre-expiry (the hedge).** Default stays 60s for all providers. X sets 45 min — wider than the 30-min keep-fresh tick — so X tokens rotate on the heartbeat ~every 90 min *before* expiry, instead of the old emergent behavior (the 60s window was essentially never hit by a tick, so tokens fully expired and were refreshed reactively, up to ~30 min late, by whatever needed them; access tokens were live only ~89% of the time). Expected effect on grant survival: possibly none — all ~650 of Sam's *successful* refreshes were also late/post-expiry, so late presentation is at most a weak risk factor. What it definitely does: access tokens live ~100% of the time, refreshes stop riding inside busy sync tasks, and the keep-fresh task finally does what its docstring says. Near-zero risk; other providers unaffected.

## Validation

New test pins the margin boundary from both sides (wide-margin provider rotates a token 30 min from expiry; default-margin provider does not). Existing single-flight concurrency tests pass unchanged. Live cohort under the new cadence with full logs: @henrytdowling's fresh grant + expected reconnects from the affected users — if a cohort grant dies, the logs will finally say what it was doing.

## Follow-up (separate PR, higher product value)

Grant-death notification: keep-fresh already detects a dead grant within 30 minutes; today that surfaces only as a banner on a page nobody visits. An email/notification turns the residual rare death into a same-day two-click reconnect instead of silent data staleness.

### DIFF
--- a/backend/integrations/storage.py
+++ b/backend/integrations/storage.py
@@ -12,6 +12,7 @@ returning it.
 
 from __future__ import annotations
 
+import logging
 from datetime import UTC, datetime, timedelta
 from uuid import UUID
 
@@ -22,6 +23,8 @@ from ..database import get_pool
 from .base import AccountInfo, TokenSet
 from .crypto import integration_fernet
 from .registry import get_provider
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_ACCOUNT_KEY = "default"
 
@@ -118,10 +121,18 @@ async def store_token(
 
 
 _TOKEN_QUERY = """
-    SELECT access_token_encrypted, refresh_token_encrypted, expires_at
+    SELECT access_token_encrypted, refresh_token_encrypted, expires_at, updated_at
     FROM user_integrations
     WHERE user_id = $1 AND provider = $2 AND account_key = $3
 """
+
+# How close to expiry a token must be before a read refreshes it. Providers
+# with single-use rotating refresh tokens (X) override `refresh_margin` wide
+# enough that the 30-min keep-fresh tick always rotates BEFORE expiry: every
+# grant death we've autopsied was a refresh token first presented well after
+# its access token expired, so rotation happens on a calm schedule instead of
+# at the moment of use.
+_DEFAULT_REFRESH_MARGIN = timedelta(seconds=60)
 
 
 def _require_token_row(row, provider: str) -> None:
@@ -136,8 +147,8 @@ def _require_token_row(row, provider: str) -> None:
         )
 
 
-def _needs_refresh(expires_at: datetime | None) -> bool:
-    return expires_at is not None and expires_at < datetime.now(UTC) + timedelta(seconds=60)
+def _needs_refresh(expires_at: datetime | None, margin: timedelta) -> bool:
+    return expires_at is not None and expires_at < datetime.now(UTC) + margin
 
 
 async def get_valid_token(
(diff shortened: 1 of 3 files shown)


henry-dowling:
**ROOT CAUSE FOUND — supersedes the attribution narrative in the description.** The celery worker's `TWITTER_OAUTH_CLIENT_ID` env var was stale (different from the backend's). Every refresh executed on the worker — including all keep-fresh ticks — presented the wrong client id and got 401 **without consuming the refresh token**; refreshes executed on the backend (triggered by live web usage) succeeded. That explains every 'grant death': none of the grants were revoked at all. Proven by presenting a 'dead' refresh token with the backend's credentials: 200 OK, fresh pair returned. Worker env corrected at 16:54 UTC; affected grants are expected to revive on the next keep-fresh tick without user action (subject to X's idle-cull for the two long-idle tokens). This PR's margin change and refresh logging remain useful (the `idle_s`/outcome log line would have exposed this pattern — worker refusals alongside backend successes — in minutes), but the description's per-grant death theories (probe anomaly, self-revocation, rare X refusal) are all obsolete: it was one stale env var.

samzliu (approved):

PR #972, opened 2026-08-06
Filter unified search by a last-modified datetime range

ColemanDuPlessie:
## Changes

Allows the user to sort by date when searching stash. This is accomplished by adding the --modified-after and --modified-before CLI options, and similarly allowing the user to select from a datetime range in the web UI. The UI is a bit vibecoded at the moment, but works.

Default behavior remains the same as before, changes only occur if the user decides to sort by date.

Finally, some integrations don't actually give us a date of last modification, and hits from those are only included if the user doesn't sort by date.

## What (Claude-generated below this point)

Adds `modified_after` / `modified_before` bounds to unified search, applied on every result path:

- Sessions, pages, and exact-ref matches are bounded in-SQL, as is the copied-content FTS path.
- Federated (connected-source) hits are post-filtered.
- Results with no last-modified date are excluded whenever either bound is set.

## Surfaces

- **CLI**: `stash search --modified-after / --modified-before`
- **MCP**: same parameters on the search tool
- **Search page**: a fourth filter chip with presets plus a custom range (`SearchModifiedFilter`)
- **Docs**: CLI docs page updated

## Tests

Backend (`test_sources.py`), CLI (`test_sources_cli.py`), and frontend (`modified-range.test.ts`, `SearchModifiedFilter.test.tsx`) coverage included.

## Notes

- Branch predates ~77 commits of main but merges cleanly (verified with `git merge-tree`).
- Screenshots of the new search-page chip to follow: the local stack's backend port is currently held by a benchmarking run, so I couldn't capture the UI yet.

### DIFF
--- a/backend/routers/sources.py
+++ b/backend/routers/sources.py
@@ -155,6 +155,8 @@ async def search_sources(
     include_sources: list[str] | None = Query(None),
     exclude_sources: list[str] | None = Query(None),
     limit: int = Query(20, ge=1, le=500),
+    modified_after: datetime | None = Query(None),
+    modified_before: datetime | None = Query(None),
     current_user: dict = Depends(get_current_user),
     scope_user_id: UUID = Depends(get_scope),
 ):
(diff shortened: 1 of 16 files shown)


henry-dowling (approved):
LGTM!

PR #980, opened 2026-08-07
fix: agent skill tools return real app URLs and never collide on names

henry-dowling:
This is complementary to the fix in #974. Like 974, but for agents- our fix for agents is to just force it to come up with a new name (or, edit the existing skill) rather than automatically renaming. We prefer this because if an agent suggests a colliding name, we're gonna have to tell it that the name collides, so at that point we may as well ask it to give us a new name. This will fix part of the jank that happened during today's incident with Skills UX- Mike's agent hallucinated a URL because we don't return URLs on skill creation via the skill tool!

## What happened

Two agent-tool gaps from the 2026-08-06 incident review:

1. **Models invent URLs when tools don't provide them.** `create_skill` returned bare `{folder_id, name}`; the customer's chat agent then composed `app.joinstash.ai/skills/{folder_id}` — which is the *published-slug* route — and the customer landed on a confident "Skill not found" for a skill that existed. `create_skill` and `list_skills` now include `app_url` pointing at the canonical `/skills/folder/{id}` page.
2. **The agent tool still had the exact-name create bug** that #974 fixed for the web button: `create_folder(name)` raw, erroring on any squatted name. It now uses the same uniquifying `create_skill` service — which gains an optional `skill_md` parameter so agents keep authoring the full SKILL.md up front (None = the starter template, named after the folder the create actually landed on).

## Tests

- Lifecycle test extended: `app_url` asserted on both create and list results.
- New: two same-named agent creates → "Fitment" and "Fitment (2)", each keeping its own authored SKILL.md body.
- Agent-runtime + skills suites pass (27 tests), ruff clean. Backend-only, no migration, no GUI change.

### DIFF
--- a/backend/services/agent_runtime.py
+++ b/backend/services/agent_runtime.py
@@ -332,6 +332,12 @@ async def _query_table(args: dict) -> dict:
     )
 
 
+def _skill_app_url(folder_id: object) -> str:
+    """The skill's page in the app. /skills/{slug} belongs to *published*
+    skills only — a folder id pasted there renders "Skill not found"."""
+    return f"{settings.PUBLIC_URL.rstrip('/')}/skills/folder/{folder_id}"
+
+
 @tool(
     "list_skills",
     "List skills (folders with SKILL.md) in this Stash account, with their "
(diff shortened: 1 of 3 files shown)


henry-dowling:
Amended after review discussion: on a taken name the tool no longer silently creates `name (2)` — the model would believe it created `Fitment` while `load_skill("Fitment")` resolves to the older skill. A collision now returns a structured refusal (`error: name_taken`, the existing holder's id and real URL, and a hint) so the model can update the existing skill or choose a new name deliberately. The `app_url` additions are unchanged; the shared service's `skill_md` parameter is reverted (no longer needed). Tests updated to pin the refusal semantics.

PR #983, opened 2026-08-07
pin markdown-it to 14.2.0 — render parity with the removed server

henry-dowling:
One-liner that missed #981's merge by a minute: the editor's client-side renderer was installed as 15.0.0 while the old server bootstrap ran 14.2.0. Pinning to the identical version removes the only render-difference risk in the switchover. Tests unchanged (276 pass).


henry-dowling:
not needed after #992, closing

PR #991, opened 2026-08-07
test: enumerate every way to create a skill

henry-dowling:
This is a test-only PR.

# Slop
## Why

A skill is two writes: mark the folder (`folders.is_skill`) and give it instructions (a `SKILL.md` page). Every creation path has to remember both. When membership moved from derived to stored in #985, **six paths silently forgot** — the three CLI commands, the MCP server tool, and both convert buttons kept writing a SKILL.md into folders that were never marked skills, producing invisible folders. Each was found by hand, hours apart, after shipping; two of them reached production.

Nothing failed. The test suite covered the paths it happened to cover, and I mistook that for coverage of the paths that exist.

## What this does

Walks every creation path that can run in-process — `create_skill`, convert-a-folder, publish, fork, bulk file write (GitHub import + skills sync), and the agent tool — and asserts both halves landed: the folder is marked, and the skill lists with `has_instructions: true`.

The CLI and MCP server create over HTTP and can't be exercised here, so they're asserted by source inspection: each must call the convert verb after writing SKILL.md. Crude, but it is exactly the check whose absence let all four break.

## Verified to fail

A test that can't fail is decoration, so I broke things on purpose:

- Removing the flag write from `create_skill` → `AssertionError: create_skill: folder was not marked a skill`
- Removing the convert call from the MCP server → `cli/mcp_server.py: expected at least 1 convert_folder_to_skill call(s), found 0`

Both restored; suite green (16 passed with the membership tests).

## The rule it encodes

Adding a new way to create a skill means adding it here. That is the whole point — the next person who forgets a half finds out from CI in eight minutes instead of from a customer in three days.

### DIFF
--- a/backend/tests/test_every_skill_creation_path.py
+++ b/backend/tests/test_every_skill_creation_path.py
@@ -1,0 +1,145 @@
+"""Every way to create a skill must produce a skill with instructions.
+
+Skill membership is a stored flag (folders.is_skill) and the instructions are
+a page named SKILL.md. Those are two separate writes, so every creation path
+has to remember both — and on 2026-08-07 six of them silently forgot: the
+three CLI commands, the MCP server tool, the folder-convert button, and the
+skill-view demote button all kept writing a SKILL.md into a folder that was
+never marked a skill. Each produced an invisible folder instead of a skill,
+and each was found by hand, hours apart.
+
+This enumerates the paths so the next one that forgets fails the build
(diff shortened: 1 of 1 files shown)


ColemanDuPlessie (approved):
Looks good, one style note that doesn't really matter is that a lot of the generated test names are really verbose, or even full sentences (e.g. test_the_client_side_paths_all_call_the_convert_verb() where I would expect something more like test_client_paths_convert_skills() or similar)

PR #993, opened 2026-08-07
Fix: five places sent a folder id to the published-skill route ("Skill not found")

henry-dowling:
`/skills/<x>` resolves a **published skill slug**. Five places handed it a folder id instead, and every one of them rendered "Skill not found".

### The five

| Where | What the user does | |
|---|---|---|
| `FolderClient.tsx:102` | Opens a skill from the **Files** tree — it lands on `/folders/<id>`, which self-heals onto the skill route, into the 404 | fixed |
| `AddToStashButton.tsx:55` | Adds someone's public skill, then clicks **Open \<name\>** — the whole Discover install path | fixed |
| `SkillFolderClient.tsx:88` | Clicks a **breadcrumb** to go back up from a subfolder | fixed |
| `SkillFolderClient.tsx:172` | Opens a **subfolder inside a skill** — so any skill with a subfolder can't be browsed past its top level | fixed |
| `SkillFolderClient.tsx:144` | **Shares a skill.** This is the path the Share dialog turns into the copied link, so the recipient got a 404 | fixed |

All five now use `/skills/folder/<id>`, matching `urlForTab`, the convert-to-skill flow, and `SkillPageClient`.

### Why it took two passes

The first pass grepped for the literal text ``/skills/${...}`` and caught two. It missed the other three because `SkillFolderClient` passes a *prefix* to `FileBrowser`, which builds the URL as `` `${folderHrefBase}/${item.id}` `` — so the broken pattern never appears at the call site. Every `/skills` URL construction in the frontend has now been enumerated and classified by hand; the remaining ~40 all pass a real slug or already use the folder route.

Two existing tests in `SkillFolderClient.test.tsx` asserted the broken URLs (`href: "/skills/folder-root"`, `data-href-base="/skills"`). They encoded the bug, so they are corrected here.

### Testing

Every assertion was confirmed to fail against the old code before the fix. The share link had no coverage at all and now does.

- **Verified live in a browser:** opening a skill from Files, and opening a subfolder inside a skill.
- **Verified by unit test:** breadcrumbs, the share link, the Discover install path.

Frontend **292 passed** (from 279), ESLint 0 errors. Two pre-existing `tsc` errors reproduce identically with these changes stashed and are untouched; CI's `next build + type-check` passes.

### DIFF
--- a/frontend/src/app/(app)/folders/[folderId]/FolderClient.test.tsx
+++ b/frontend/src/app/(app)/folders/[folderId]/FolderClient.test.tsx
@@ -1,0 +1,108 @@
+import { cleanup, render, waitFor } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import FolderClient from "./FolderClient";
+import { getFolderContents } from "@/lib/api";
+
+const router = vi.hoisted(() => ({
+  push: vi.fn(),
+  replace: vi.fn(),
+}));
+
+vi.mock("next/navigation", () => ({
+  useParams: () => ({ folderId: "folder-root" }),
+  useRouter: () => router,
+  useSearchParams: () => new URLSearchParams(),
+}));
+
+vi.mock("@/lib/api", () => ({
+  ApiError: class ApiError extends Error {
+    status = 500;
+  },
+  convertFolderToSkill: vi.fn(),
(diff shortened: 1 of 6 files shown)


henry-dowling:
Screenshots

samzliu (approved):

PR #1031, opened 2026-08-10
Make /files and /sessions fast: session listings read the session row

henry-dowling:
`/files` and `/sessions` were slow in a way that scaled with how much a user had stored. Measured in prod on one real account (**4,346 sessions, 201,680 events, 4.9GB of `history_events`**):

| endpoint | before | payload |
|---|---|---|
| `/me/sources/tree?depth=4` | **18.2s / 19.1s** | 44KB |
| `/me/sidebar` | **10.8s / 6.4s** | 370–488KB |
| `/me/sessions?limit=200` | **8.2s / 5.4s** | 19KB |

19 seconds for 44KB means the cost is work, not bytes. Three causes, each fixed.

## 1. Listing sessions aggregated every event the user owned

`LIMIT 200` applied *after* a `GROUP BY` over all of `history_events`, and a CTE read `content` for every event to name the few sessions without a stored title:

```
Seq Scan on history_events   rows=201,683, removed by filter 636,297
Sort Method: external merge  Disk: 49496kB  (x3 workers)
Buffers: shared read=580,764                (~4.5GB)
Execution Time: 6252 ms
```

That title fallback serves **8%** of sessions — 8,136 of 8,860 already have a row in `session_titles`. It now belongs to the title service, which fetches the first message only for sessions that lack a stored title, keyed by `session_id` so it rides `idx_history_events_workspace_session_created`.

Then the aggregate itself went away. How many events, how big, and when the last one landed are known at write time, so **migration 0185** stores them on the session row and the listing became an indexed read of a 10MB table:

| `/me/sessions?limit=200` | time | buffers |
|---|---|---|
| before | 6,252ms | 580,764 |
| after intermediate fix (no title scan) | 377ms | 122,901 |
| **after materialization** | **7ms** | **555** |

**~900× faster, 1,000× less I/O.** (7ms measured without the new index, using `started_at` as a stand-in; the shipped index is `(owner_user_id, last_event_at DESC)`.)

## 2. Source trees walked providers one at a time

`sources_tree` awaited each provider in sequence, so a scope cost the *sum* of its connections — nine providers, nine serial walks, each pulling up to `TREE_DOC_LIMIT` rows. Providers and their members now resolve concurrently, and the tree's page/session queries share one gather.

## 3. The tree named sessions from raw transcript content

Session labels come from `session_titles`, so the VFS listing is no longer a reason to read the event table either.

## Keeping the roll-up true

`session_service.record_event_stats` folds each write into the session row at the single choke point both event writers already funnel through (`push_event`, `push_events_batch`). Verified against a live recompute — count, compressed size and last timestamp match exactly. `size_bytes` stays `pg_column_size` (the stored, compressed size the old aggregate reported) so numbers already on screen don't move, and `last_event_at` takes the greatest timestamp so importing an old transcript can't restamp a session as recent.

The migration also:
- creates session rows for the **3,478 events whose session existed only as events** — already listed on `/sessions` (LEFT JOIN) but invisible to the sidebar (INNER JOIN), so materializing without them would have dropped sessions from one view;
- backfills `created_by` where missing, because the author used to be read off the first event.

## Two regressions the tests caught, and what they taught

- **Empty session shells.** A session row with no events was naturally hidden when listings were event-driven. Reading `sessions` directly surfaced it — caught by `test_empty_session_shell_is_hidden_from_default_views`. The listings now filter `event_count > 0`, which states that rule instead of relying on a join to imply it.
- **A fixture writing behind the service.** `test_permissions` inserted events straight into the table, bypassing the roll-up, and its session vanished. The fixture now writes through the same path — a session the app could never produce isn't worth asserting on.

## Verified

- Prod `EXPLAIN (ANALYZE, BUFFERS)` for every query shape, before and after (read-only).
- Roll-up correctness asserted against a live recompute on a real write path.
- Migration applied locally: backfill populated, all three endpoints 200 with correct titles, `/sessions` and `/files` render.
- **1,366 backend tests pass**; ruff clean.

## Still open

- The sidebar returns every session (370–488KB). Capping it needs a search endpoint first, because the command palette indexes that list.
- `TREE_DOC_LIMIT = 5000` rows per source to render a 50-per-directory preview is more than the view needs.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit ab507d982596bed8a2433886038c13704b5fc6f3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


henry-dowling:
closing this-  ui shifting too much rn, we can re-implement later if needed.

PR #1035, opened 2026-08-12
Serve the Stash tools over HTTPS as a remote MCP endpoint so that we can support claude cowork.

henry-dowling:
Two changes. The first is a bug fix that stands alone; the second is the foundation for reaching Claude surfaces that aren't a developer's laptop.

## Hold mcp at 1.x (9694039d)

`mcp` 2.0.0 shipped on 2026-07-28, moved `FastMCP` to `mcp.server.mcpserver`, and dropped `streamablehttp_client`. Both of our pins were an unbounded `>=1.23.0`, so **every fresh install has been resolving to 2.0 and `stash-mcp` dies on its import line.** Anyone who installed or upgraded the CLI in the last two weeks has a dead MCP entrypoint.

Prod is not affected — its running image is on 1.x, confirmed from a live traceback that still shows `streamablehttp_client` present. But `backend/requirements.txt` carried the same unbounded floor, so the next fresh Docker resolve would have taken the Granola and PostHog clients down at import time. That tripwire is the main reason this is worth shipping on its own.

Pinning rather than porting: the CLI half is a two-line change (`MCPServer` is a drop-in), but the backend half needs a real port to the new client API, and splitting mcp majors across one repo is worse than holding both at 1.x.

## Remote MCP endpoint (72ac610c)

`cli/mcp_server.py` runs on the user's laptop and reads their API key from `~/.stash/config.json`. That works for Claude Code and cannot work anywhere else — I verified that a Cowork cloud session runs in a Linux container with no view of the user's Mac at all, and Claude on the web and mobile never touch the machine. Those surfaces reach tools only over HTTP with OAuth.

**Tools call our own REST API in process**, through an ASGI transport carrying the caller's bearer token. Authorization, scoping, plan gates and rate limits are then enforced by exactly the code that enforces them for the web app. Re-implementing 72 tools against the service layer would mean a second authorization path to keep correct, and the first missed scope check is a data leak between users. This is the main design decision in the PR and the one most worth arguing with.

**The OAuth audience is the endpoint's own URL**, not the web app's API identifier, so MCP tokens and web-app tokens are separate credentials and neither is accepted where the other is expected.

### Deployment

Gated on `MCP_PUBLIC_URL`. Unset, nothing mounts — **merging this changes nothing in production** until that variable is set on Render. Self-hosted deployments, which have no Auth0 tenant to verify tokens against, never serve it.

Requires the Auth0 tenant's *Resource Parameter Compatibility Profile* and *Include Issuer in Authorization Responses* toggles, both now enabled. Without them Auth0 ignores the `resource` parameter Claude sends and every call 401s. Existing web-app sign-ins are unchanged: they always pass `audience` explicitly, which still takes precedence.

### Scope

Three tools, to prove the path end to end. The remaining 69 are mechanical — a docstring and one line calling an existing route each — and should follow only after this has been verified against a real Claude connection, where the remaining unknowns actually live.

## Verification

The full discovery handshake works as Anthropic's spec describes:

```
POST /mcp  ->  401
   WWW-Authenticate: Bearer resource_metadata=".../.well-known/oauth-protected-resource/mcp"
GET that URL  ->  200
   {"resource": ".../mcp", "authorization_servers": ["https://<tenant>/"]}
```

- 1360 backend tests pass; ruff clean
- 7 new tests cover the discovery chain and the audience separation
- `test_activity.py` is excluded: it segfaults inside torch/sentence-transformers on my machine, which I confirmed also happens on a clean tree

## Not done

- 69 remaining tools
- Security review — this puts 72 tools, including destructive ones, on the public internet
- `mcp-review@anthropic.com` for Anthropic-held client credentials; DCR otherwise registers a new OAuth client on every fresh connection

---

> [!NOTE]
> **High Risk**
> Introduces a public OAuth-authenticated MCP surface and changes Auth0 audience validation, which are security-critical paths. Gated behind MCP_PUBLIC_URL, but once enabled it exposes tools on the public internet.
> 
> **Overview**
> Exposes Stash tools over HTTPS with OAuth so Claude surfaces that aren't a local laptop (Cowork, web, mobile) can reach them. Also **pins `mcp` to 1.x** so fresh installs stop resolving to 2.0 and breaking `stash-mcp` / Granola / PostHog imports.
> 
> The new remote endpoint mounts only when `MCP_PUBLIC_URL` is set. Tools are thin wrappers that call existing REST routes **in-process** with the caller's bearer token, so auth, scoping, plan gates, and rate limits stay on the same path as the web app. Starts with three tools: `stash_search`, `stash_vfs`, and `stash_list_workspaces`.
> 
> Auth0 JWT validation is split so MCP tokens use the endpoint URL as audience and web-app tokens keep `AUTH0_AUDIENCE` — neither credential is accepted on the other surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72ac610c7fd16a98d933f9ee47c5e9f9c6cbb04e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -164,6 +164,15 @@ class Settings:
         AUTH0_ENABLED,
     )
 
+    # --- Remote MCP endpoint ---
+    # The canonical URL Claude connects to, e.g. https://api.joinstash.ai/mcp.
+    # It is also the OAuth audience: Claude sends this URL as the `resource`
+    # parameter, and Auth0 stamps it into the token's `aud`. It must match what
+    # the user types into Claude exactly, so it cannot be derived — an
+    # almost-right URL fails discovery with an unhelpful error. Unset means the
+    # endpoint is not mounted at all.
+    MCP_PUBLIC_URL: str = os.getenv("MCP_PUBLIC_URL", "")
+
     # --- Embeddings ---
     # Provider: "openai", "huggingface", "local", or "auto" (default).
     # Auto-detect: OPENAI_API_KEY → openai, HF_TOKEN → huggingface, else → local.
--- a/backend/main.py
+++ b/backend/main.py
@@ -170,6 +170,12 @@ if settings.AUTH0_ENABLED:
 
     app.include_router(auth0_router)
 
+# Remote MCP endpoint. Mounted last: its tools call back into the routes above
+# through an in-process transport, so they all have to be registered first.
+from backend import mcp_remote  # noqa: E402
+
+mcp_remote.attach(app)
+
 
 @app.middleware("http")
 async def add_security_headers(request: Request, call_next):
--- a/backend/managed/auth0/jwt.py
+++ b/backend/managed/auth0/jwt.py
@@ -29,8 +29,25 @@ async def _fetch_jwks() -> dict:
 
 
 async def validate_auth0_token(token: str) -> dict:
-    """Validate an Auth0 access token. Returns the decoded claims."""
-    if not settings.AUTH0_DOMAIN or not settings.AUTH0_AUDIENCE:
+    """Validate an Auth0 access token minted for the web app's API."""
+    if not settings.AUTH0_AUDIENCE:
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Auth0 not configured",
+        )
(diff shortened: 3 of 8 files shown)


henry-dowling:
closing because not high prio right now

PR #1037, opened 2026-08-12
The curator maintains three skills alongside the wiki

henry-dowling:
The Memory wiki is **pull**: it helps only when an agent thinks to search it. Skills are **push** — the harness loads every SKILL.md description at session start — so a curated skill fires without the agent knowing to look. This teaches the nightly curator to compile both from the same run.

## Three slots, enforced in the service

Every invariant lives in `curated_skill_service`, not in the prompt, so a prompt regression degrades the skills' quality instead of reintroducing the skill sprawl the cap exists to prevent.

- **Three slots.** A fourth write 400s and must name the slot it replaces; the replaced skill is deleted.
- **A human edit adopts.** Editing a page inside the folder, pushing contents through sync, publishing, or demoting to a plain folder each clear the curated flag. The slot frees, and the curator never touches that skill again.
- **Adoption is permanent.** Writing a name that belongs to a non-curated skill is refused, so the curator can't resurrect its version under that name or a uniquified `Name (2)` competing for the same trigger.

`folders.is_curated` carries a CHECK that a curated folder is always a skill, which inherits the existing `folders_protected_is_never_a_skill` constraint — Memory and Clips can't be captured.

## Sync: two branches, not a new codepath

Curated skills ride the normal three-way sync. Only two things differ:

- A curated skill whose local copy changed **pushes instead of conflicting**. The push adopts it server-side, so there's no second author left to disagree with.
- A dir **sync itself pulled** is deleted when its cloud copy is gone, so a skill the curator retired stops loading. Keyed on `rec` as well as frontmatter — frontmatter is copyable, so it can't be the sole license to delete a directory.

## Corrections become their own feed input

The change feed drains oldest-first through the watermark, and tool events outnumber user messages by ~50:1. Measured on a real account: **500 feed events held 295 tool events and 6 user messages.** The words a skill is mined from lose every cap they compete for.

`corrections` is a rolling 30-day window sampled per session, deliberately outside the watermark — it re-presents the same messages every night, because recurrence across sessions is the test for a skill and one night's delta can't show it. The wiki's delta semantics are untouched.

## Tests

1378 backend, 237 CLI, 309 frontend vitest; ruff and eslint clean. New coverage locks the cap, all four adoption paths, eviction, the sync branches, and the corrections window (including that one chatty session can't crowd out the others).

## What is NOT proven

**The full pipeline has never run.** Feed → prompt → three slots has not been exercised by a real curator run; the mechanism is tested, the output is not.

An offline experiment gave the rendered prompt and a real corrections window to a fresh agent. The prompt's discipline held — it counted distinct sessions per candidate, rejected four for failing recurrence, and left the third slot empty rather than filling it. But its two picks came from short lookup sessions, because **of the 120 sessions the sampler selects, 67 have only one or two user messages.** A 191-turn working session contributes the same four messages as "what time is my flight."

So session weighting is an open problem: the sampler maximizes session count, which is what the recurrence rule counts, but that gives a throwaway ask the same standing as a day of engineering. Worth resolving before this is trusted to write skills unattended.

**Suggested rollout:** merging deploys to prod, and the nightly curator would then write skills into every user's Stash, which sync pushes into their agents' skills dirs. Consider gating to one account until a few real runs have been watched.

---

> [!NOTE]
> **Medium Risk**
> Adds curator-owned skills that load into every agent session and changes skills sync (local retire/delete, conflict bypass). Cap and adoption are server-enforced, but a bad curator run would still write into users' skills dirs.
> 
> **Overview**
> Enables the nightly curator to maintain **at most three push skills** alongside the Memory wiki. Skills load into every agent session; the three-slot cap, eviction, and adoption rules live in `curated_skill_service` so a prompt regression can't sprawl.
> 
> **Adoption is permanent:** editing a page in the folder, sync push, publishing, or demoting to a folder clears `is_curated` and frees the slot. The curator cannot rewrite an adopted name.
> 
> Adds a **`corrections` feed** — a rolling 30-day window of user messages sampled per session, outside the watermark — so skills are mined from recurrence across sessions rather than competing with tool traffic in the delta. The curator prompt and `stash skills curated` CLI teach/write against that input.
> 
> Skills sync treats curated skills specially: a local edit **pushes instead of conflicting** (and adopts), and a retired curated skill sync itself pulled is **deleted locally** so it stops loading. The Brain dashboard surfaces the three slots for review.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbe588f9bbe6ec01edf219bda79d9c08804e7618. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/migrations/versions/0185_curated_skills.py
+++ b/backend/migrations/versions/0185_curated_skills.py
@@ -1,0 +1,39 @@
+"""AI-curated skills: three reserved slots the nightly curator owns.
+
+The curator already compiles the user's activity into the Memory wiki. The
+wiki is pull — it helps only when an agent thinks to search it. Skills are
+push: every SKILL.md description is loaded at session start, so a curated
+skill fires without the agent knowing to look.
+
+Curated skills are ordinary skill folders (same ``is_skill`` flag, same
+sync, same Skills surface) marked ``is_curated`` so two things are
+expressible: the three-slot cap, and adoption — a human edit clears the flag
+and the curator may never touch that skill again.
+
+Revision ID: 0185
+Revises: 0184
+"""
+
+from alembic import op
+
+revision = "0185"
+down_revision = "0184"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE folders ADD COLUMN is_curated boolean NOT NULL DEFAULT false")
+    # A curated folder is always a skill, so the existing
+    # folders_protected_is_never_a_skill check keeps Memory and Clips out too.
+    op.execute(
+        "ALTER TABLE folders ADD CONSTRAINT folders_curated_is_a_skill "
+        "CHECK (NOT is_curated OR is_skill)"
+    )
+    op.execute("CREATE INDEX idx_folders_is_curated ON folders(owner_user_id) WHERE is_curated")
+
+
+def downgrade() -> None:
+    op.execute("DROP INDEX IF EXISTS idx_folders_is_curated")
+    op.execute("ALTER TABLE folders DROP CONSTRAINT IF EXISTS folders_curated_is_a_skill")
+    op.execute("ALTER TABLE folders DROP COLUMN IF EXISTS is_curated")
--- a/backend/routers/files_tree.py
+++ b/backend/routers/files_tree.py
@@ -31,6 +31,7 @@ from ..models import (
 )
 from ..services import (
     comment_service,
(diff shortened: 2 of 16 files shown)


henry-dowling:
let's re-implement this once SDP + External multiplayer is done. That will make it much easier to get the shape of skill distillation right.

PR #1039, opened 2026-08-12
The CLI can authorize a provider, not just index one

henry-dowling:
`stash sources add` indexes a mailbox or repo on an OAuth grant that already exists. Nothing in the CLI could *create* that grant, so the first half of connecting a source had no agent path at all — the user had to be told to go find the integrations page and click through it. Trying to add a personal Gmail account failed at `sources add` with a 400 that never mentioned the step it was missing.

## What this adds

`stash sources connect <provider>` — fetch the provider's authorize URL, open it, watch the account list until the grant lands, fail loud if it never does.

The consent itself stays the user's: it wants a password and 2FA, and an agent holding those is the thing OAuth exists to prevent. What was missing is the handoff around it. Both halves already existed — the backend returns an authorize URL for the frontend (it can't 302 with a Bearer header), and the CLI already had the open-a-URL-then-poll shape in its own sign-in flow. This joins them.

## Detecting that the grant landed

A new `account_key` is not enough. Repairing an expired grant reuses the key, so on key alone a finished reconnect is indistinguishable from an abandoned one, and the command would wait out its whole timeout on a consent that actually succeeded. Account status now exposes `updated_at`, which every `store_token` rewrites, and the poll fingerprints on `(account_key, updated_at)` over non-disconnected accounts.

## Error messages

Both Gmail 400s now name the authorized mailboxes. "Gmail account is not connected" is equally true whether the mailbox was never authorized or the address was misspelled, and those want opposite fixes.

## Notes for review

- `_is_ssh()` is extracted from `_browser_auth_flow` so sign-in and this flow agree on when a browser can be opened; behavior there is unchanged.
- Under `--json` the authorize URL goes to **stderr** so it can still reach a human without corrupting the payload on stdout.
- `updated_at` is additive on the integrations API; the frontend ignores unknown fields and its tests pass unchanged.
- One existing assertion in `test_gmail_multi_account.py` pinned the old error string and now asserts the mailboxes are named.

## Testing

- 7 new CLI tests, 2 new backend tests, 1 added to `test_gmail_multi_account.py`.
- Mutation-checked the fingerprint: keying on `account_key` alone fails exactly the reconnect test, with the real symptom ("Timed out… Nothing was connected").
- Full CLI suite (241) plus `test_sources`, `test_integration_sources`, `test_gmail_multi_account`, `test_integration_connect_signal`, `test_disconnect_retention`, `test_integration_reconnect` (168) green; ruff check + format clean.
- Exercised end-to-end against prod: authorized a second Gmail account, indexed it, synced, and searched it successfully.

---

> [!NOTE]
> **Medium Risk**
> Touches the OAuth connect handoff and integrations status API, but reuses existing authorize/status endpoints and only adds an additive `updated_at` field. No changes to token exchange or storage security.
> 
> **Overview**
> Adds **`stash sources connect <provider>`**, which opens the provider consent URL and polls until the OAuth grant lands — the missing step before `sources add`.
> 
> Account status now exposes **`updated_at`**, so the CLI can detect a reconnect that reuses the same `account_key`. Polling fingerprints on `(account_key, updated_at)` over live accounts and fails loudly on timeout.
> 
> Gmail add errors now **name the authorized mailboxes**, so callers can tell “authorize first” from “pick which one” or a typo.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca789759bcd906a8368d4564942945d2bfa9acc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/integrations/storage.py
+++ b/backend/integrations/storage.py
@@ -57,6 +57,11 @@ def _account_row(row) -> dict:
         "scopes": list(row["scopes"] or []),
         "expires_at": row["expires_at"].isoformat() if row["expires_at"] else None,
         "connected_at": row["created_at"].isoformat() if row["created_at"] else None,
+        # Every store_token rewrites this, so a re-authorization of an account
+        # that already exists moves it. `stash sources connect` polls on it to
+        # tell "the browser consent finished" from "nothing happened yet" —
+        # account_key alone can't, since a repair reuses the same key.
+        "updated_at": row["updated_at"].isoformat() if row["updated_at"] else None,
     }
 
 
@@ -311,7 +316,7 @@ async def status(user_id: UUID, provider: str) -> dict:
     rows = await pool.fetch(
         """
         SELECT account_key, scopes, expires_at, account_email, account_display_name,
-               created_at, access_token_encrypted IS NULL AS disconnected
+               created_at, updated_at, access_token_encrypted IS NULL AS disconnected
         FROM user_integrations WHERE user_id = $1 AND provider = $2
         ORDER BY account_email NULLS LAST, account_display_name NULLS LAST, account_key
         """,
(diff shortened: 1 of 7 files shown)


henry-dowling:
closing for now, not absolutely crucial to our roadmap

PR #1040, opened 2026-08-12
Allow a Google Drive folder to provide Skills

henry-dowling:
(1) You can now use a google drive document as content for a skill. You just need to format the google doc like a skill (eg use the standard skill yaml frontmatter)

CleanShot 2026-08-18 at 00 18 04@2x

and then mark the folder as "skill" in our UI — either with **Use for Skills** on the folder's integration page, or by pasting the folder's Drive link into the Skills page's **Add external skill by link** box

(2) It's no longer possible to share a `source` via stash. Eg, you can't share your google drive with another user via stash. We think this is a good product decision because it was always a little weird that you could do that (why not just share it in drive?) and removing this now eliminates a class of bug with this PR where it wasn't possible to recognize a shared google drive document as a skill

(3) we also change the way URLs for skills are constructed so that they do not rely on the name of the skill anymore. More precisely, we previously derived skill URL from our postgres ID for the skill which was itself keyed on (source_id, path), and "path" contained the name of the skill. This is necessary because this PR invites the possibility that a shared skill in google docs could change its name, which would cause us to need to update all references to it which would be annoying.

---

> [!NOTE]
> **Medium Risk**
> Touches agent skill loading, Drive sync/deletion semantics, and removes delegated source sharing (breaking for anyone relying on shared integrations); incorrect folder-empty handling was high-impact but is explicitly guarded.
> 
> **Overview**
> **Google Drive folders can act as skill shelves** without copying content into Stash. A new `user_sources.binds_skills` flag (and `bind`/`unbind` APIs) marks a picked `google_drive_folder` source; documents in `drive_documents` that pass the same SKILL.md frontmatter validation as folder skills are listed, read, and exposed to agents as `backing: "source"` skills addressed by **upstream `external_ref`** (stable across Drive renames). Folder sync is tightened (5‑minute interval for bound folders), Google Doc markdown export is **unescaped** so frontmatter parses, and a **pre-walk folder readability check** stops sync instead of treating a revoked/trashed folder as empty (which would mass-delete indexed docs).
> 
> **Product and access changes:** connected **sources can no longer be shared** (DB cleanup + removal from share/permission paths); connected source listing is **owner-only**. The integrations and Skills UIs add “Use for Skills,” per-file skill status in bound folders, Drive-folder paste-to-connect on the Skills page, read-only `/skills/source/{ref}` pages, re-sync, and draft handling; **Run** from the Skills grid is removed. Agents get `list_skills`/`read_skill` updates (shelf name, source URLs, ambiguity when names collide). Published-skill install no longer treats a same-named source-backed doc as “already held.”
> 
> **Supporting:** `mcp` pinned to `<2`; optional Drive folder naming on source create when display name omitted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf89be0e6bd45e44260da8b304832a646da523bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ColemanDuPlessie (approved):
Looks good! One note is that there are a lot of functions where you've edited the docstring but not the content. I'm like 90% sure this is because the functionality of the functions that they call changed, so it's fine, but possibly worth noting. Also, it seems like a bit of the UI text in this change is visibly Claude-written, in a way that shows up to customers and might become grating. It's not a lot though, and it doesn't impact functionality at all, so I assume that's fine.

PR #1043, opened 2026-08-12
Address a source-backed skill by its upstream Drive file

henry-dowling:
**Draft. Stacked on #1040** — base is `feat/source-backed-skills`, so review that first.

## The bug

`drive_documents` is keyed `UNIQUE (source_id, path)` and upserted by path. Renaming a document in Drive therefore looks like *"old file gone, new file arrived"*: the row is deleted and a new one inserted, even though Drive's own file id never changed.

Reproduced before fixing:

```
same Drive file, renamed by its author:
  skill id before: fabcfa6f-…
  skill id after:  79178608-…
  identity survived: False
```

For ordinary source documents that was invisible. #1040 makes it user-visible: a skill's page, its pin, its VFS path, and **any link an agent was handed** all point at the row. So renaming a Doc silently 404s every reference to that skill.

## The fix

Skills are addressed by `source_ref` — the upstream file's own id — instead of our row id. That survives the rename. It's also the id in the document's own Drive URL, so a skill's address and the Doc's address agree:

```
/skills/source/1saZbRU77h4CNqt4Cc7vCbwphhKP4qtZg5iwBZJLHvw8
docs.google.com/document/d/1saZbRU77h4CNqt4Cc7vCbwphhKP4qtZg5iwBZJLHvw8/edit
```

`source_doc_id` is renamed to `source_ref` across the service, router, agent tools, VFS projection, and frontend, and the route segment becomes `/skills/source/[sourceRef]`. The rename is mechanical — the discriminated union on `Skill` meant the compiler found every frontend site.

## The one collision, accepted deliberately

`_resolve_shortcut` gives a Drive shortcut its **target's** file id. A shortcut sitting beside its target in the same folder therefore puts one id on two rows. They are the same upstream document, so either answers the read; the query orders by path so it is at least the same one every time. Documented at the call site.

## Why not re-key the table

Making `drive_documents` itself key on `external_ref` would fix the churn at the source — no delete-and-reinsert, no re-extraction (an OCR pass) for a rename. It's also a migration on a live content table that changes behaviour for **every** Drive folder source, not just skills, so it wants its own PR and its own review.

## Test plan

- New test: a rename upstream leaves the skill's address unchanged and still readable.
- Existing source-backed tests moved onto the new address.
- 190 backend tests pass across the affected suites; `tsc`, vitest (311), ruff and eslint clean.
- Verified end to end against the live Drive folder — app page and `stash vfs ls /skills` both serve the skill at the Doc's own id.

---

> [!NOTE]
> **Medium Risk**
> Touches skill identity, API URLs, pins, and agent links end-to-end; shortcut duplicate `external_ref` is handled but not eliminated. Existing bookmarks using old internal doc ids will 404 until users follow new links.
> 
> **Overview**
> **Fixes broken skill URLs after a Drive rename** by keying source-backed skills on the upstream file id (`external_ref` / `source_ref`) instead of the internal `drive_documents` row id, which is recreated when path-based sync treats a rename as delete+insert.
> 
> The API route becomes `GET /api/v1/me/source-skills/{source_ref}`; listings and reads filter on `external_ref`, skip docs without one, and use `ORDER BY d.path LIMIT 1` when shortcut rows share the same id. **`source_doc_id` is renamed to `source_ref`** across `skill_service`, agent `list_skills` / `read_skill` URLs, VFS `/skills`, and the frontend (`/skills/source/[sourceRef]` with `encodeURIComponent`).
> 
> A new backend test asserts the skill address stays readable across a simulated rename.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50dc72710fe638fc6b87ac0038f7039b770a3e84. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/routers/skills.py
+++ b/backend/routers/skills.py
@@ -225,16 +225,17 @@ async def get_local_skill(
     return skill
 
 
-@me_router.get("/source-skills/{doc_id}")
+@me_router.get("/source-skills/{source_ref}")
 async def read_source_skill(
-    doc_id: UUID,
+    source_ref: str,
     current_user: dict = Depends(get_current_user),
     owner_user_id: UUID = Depends(get_scope),
 ):
-    """Read one source-backed skill by the document that backs it. Addressed by
-    document rather than name because a shelf can hold two docs with the same
-    title, and the page has to open the one that was clicked."""
-    skill = await skill_service.read_source_skill(owner_user_id, doc_id, current_user["id"])
+    """Read one source-backed skill by the upstream file that backs it.
+    Addressed by file rather than name because a shelf can hold two documents
+    with the same title, and by the upstream id rather than ours because a
+    rename in Drive replaces our row."""
+    skill = await skill_service.read_source_skill(owner_user_id, source_ref, current_user["id"])
     if not skill:
         raise HTTPException(status_code=404, detail="Skill not found")
     return skill
--- a/backend/services/agent_runtime.py
+++ b/backend/services/agent_runtime.py
@@ -338,11 +338,12 @@ def _skill_app_url(folder_id: object) -> str:
     return f"{settings.PUBLIC_URL.rstrip('/')}/skills/folder/{folder_id}"
 
 
-def _source_skill_app_url(doc_id: object) -> str:
-    """A source-backed skill has no folder, so it has its own read-only page.
+def _source_skill_app_url(source_ref: object) -> str:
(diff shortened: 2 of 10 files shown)


henry-dowling:
we pulled this into #1040

PR #1057, opened 2026-08-17
stash sql reads mounted external Postgres (Supabase et al.)

henry-dowling:
this is a bigger guy. I'm gonna ignore this one for now since it's not immediately needed by anyone.

## What

A scope can mount a customer-side Postgres and query it through `stash sql`:

```
stash mounts add supabase 'postgresql://<read-only-role>@db.<project>.supabase.co:5432/postgres'
stash sql "SELECT o.vin, o.observation, p.tool_match
           FROM supabase.world_observations o
           JOIN supabase.reference_rows p USING (part_number)"
```

Every table in the mount's remote schema that the role can SELECT materializes into the per-query DuckDB under a schema named after the mount — joinable against native tables, visible in `information_schema`, bare-name views when unique. 

## Trust & safety posture

- **The customer's read-only role/view IS the access control** (same caller-side trust contract as org-scoped uploads). We additionally open every remote session with `default_transaction_read_only = on` plus a 10s statement timeout, so a mis-granted role still can't be written through and a wedged remote fails the query instead of hanging the worker.
- **User SQL still has zero network access.** Rows are fetched server-side by `pg_mount_service`; DuckDB keeps `enable_external_access` off. This is not `ATTACH` from user SQL.
- **DSNs are Fernet-encrypted at rest** (same keyring as OAuth tokens) and never returned by any endpoint after creation — list/create responses carry host + database only.
- **Count-before-fetch budget**: remote rows share the existing 200k materialization cap and are counted before fetching, so an oversized remote table fails the budget check loudly instead of OOMing the worker.
- **Unreachable mount = failed query**, not a silently smaller schema — a missing table would read to the agent as "the data doesn't exist".

## Surface

- Migration `0185_pg_mounts` (⚠️ numbering: the open Drive-skills stack and #1055 also claim 0185 — whichever lands second renumbers)
- `GET/POST /api/v1/me/sql/mounts`, `DELETE /api/v1/me/sql/mounts/{name}`
- CLI: `stash mounts add|list|remove`; `stash sql` help mentions mounts
- Mount names are validated identifiers with `files`/`memory`/`main`/system schemas reserved

## Tests

10 new tests in `test_sql_mounts.py` — the "remote" is the test database itself scoped to a fixture schema, exercising the real asyncpg path: queryable by mount schema, bare-name resolution, native↔mounted joins, numeric/jsonb type crossing, truthful `information_schema`, DSN redaction, loud failure on bad DSN / reserved + duplicate names / oversized mounts / deleted mounts, and the write-rejection guard. Existing `test_sql.py` (9) untouched and green; CLI suite 234 green; ruff clean. Full backend suite run in progress locally; CI is the gate.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit f6d043559b9659443d22b9a10c6b6c4b9b35fa9b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/migrations/versions/0185_pg_mounts.py
+++ b/backend/migrations/versions/0185_pg_mounts.py
@@ -1,0 +1,37 @@
+"""External read-only Postgres mounts, queryable through stash sql.
+
+A mount names a customer-side database (e.g. a Supabase project) whose
+tables materialize into the per-query DuckDB alongside the scope's native
+tables. The DSN is Fernet-encrypted with the integrations keyring, same as
+OAuth tokens.
+
+Revision ID: 0185
+Revises: 0184
+"""
+
+from alembic import op
+
+revision = "0185"
+down_revision = "0184"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        CREATE TABLE pg_mounts (
+            id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+            owner_user_id uuid NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+            name text NOT NULL,
+            dsn_encrypted bytea NOT NULL,
+            remote_schema text NOT NULL DEFAULT 'public',
+            created_at timestamptz NOT NULL DEFAULT now(),
+            UNIQUE (owner_user_id, name)
+        )
+        """
+    )
+
+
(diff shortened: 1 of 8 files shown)


henry-dowling:
Closing this for now because not urgent and will quickly go stale

PR #1058, opened 2026-08-17
fix: transcript events default to the whole session, not the first 100

ColemanDuPlessie:
This is a high-priority bug that significantly affects model performance: API calls that pull sessions default to only displaying the first 100 events. This is fine for the paginated web UI, but it means that, when using stash vfs to, for example, cat a file, we only actually look at the first 100 events in a transcript, and the rest are silently hidden. This PR just changes the default API behavior to be returning entire session transcripts. The web UI is still paginated, but this change lets models actually access all of long session transcripts.

## The bug

`GET /api/v1/me/transcripts/{session_id}/events` defaulted to `limit=100`.

The VFS renders `sessions/<name>/transcript.md` and `events.json` by calling that route **in one shot** (`stashvfs/model.py`), and neither client passes a limit — not `cli/client.py` (`stash vfs`), not `backend/services/vfs_service.py` (the `/me/vfs` endpoint, ask-the-stash, the `stash_vfs` MCP tool). So every session longer than 100 turns was served truncated, with no marker and exit 0.

Two ways that bites:

- **The render is oldest-first**, so the part silently dropped is the most recent work — usually the part you opened the file for.
- **`stash vfs "rg ..."` over `/sessions` returns a clean miss** for anything past the cap. The shell prefetches file bodies and regexes them, so a miss is indistinguishable from "not in any session."

It also makes two surfaces disagree. Server-side search reads `history_events` directly and is uncapped, so `stash search` can legitimately return a hit at event #400 — and a follow-up `cat` of that same session won't show it. That reads as a stale index or a hallucinated result, not as pagination.

`transcript.jsonl` in the same directory was always complete (it goes through `export.jsonl` → `read_session_events`, unbounded), which is what made the truncation easy to miss.

## The fix

Paging is now opt-in: omitting `limit` returns the whole session.

- `backend/routers/transcripts.py` — `limit: int = 100` → `limit: int | None = None`
- `backend/services/memory_service.py` — `read_session_events_page` takes `int | None`

Postgres treats `LIMIT NULL` as `LIMIT ALL`, so the unpaged read stays on the same single codepath — no branch, no second query, no compatibility shim.

**The viewer is unaffected.** `getSessionEventsPage` (`frontend/src/lib/api.ts`) always writes `limit` into the query string, so infinite scroll behaves exactly as before. The callers that *cannot* page are precisely the ones reading this route as a file, which is why the default belongs on their side.

## Tests

- `test_events_endpoint_without_limit_returns_whole_session` — 150 turns, no `limit`, asserts all 150 return with `has_more` false.
- `test_transcript_md_renders_the_whole_session` — end-to-end through the real VFS: uploads 150 turns, locates `transcript.md` via `find`, cats it, asserts the last turn is present.

Note on the second one: my first draft used `grep -r ... /sessions` and **passed even with the bug present** — it was matching the uncapped `transcript.jsonl` beside it and never touching the markdown. It now reads `transcript.md` by its own path, and a comment records why, so nobody re-broadens it.

Both tests were confirmed to fail against the old default before being committed.

## Verification

- Full backend suite: **1366 passed** (exit 0)
- `ruff check backend/ cli/` + `ruff format --check backend/ cli/`: clean

### DIFF
--- a/backend/routers/transcripts.py
+++ b/backend/routers/transcripts.py
@@ -250,14 +250,19 @@ async def get_transcript_metadata(
 @router.get("/{session_id}/events")
 async def get_transcript_events(
     session_id: str,
-    limit: int = 100,
+    limit: int | None = None,
     offset: int = 0,
     current_user: dict = Depends(get_current_user),
 ):
-    """One page of chat-thread turns for a session, oldest first, sourced
-    directly from history_events. The viewer loads the first page on open and
-    fetches more as the reader scrolls. offset is a turn ordinal, so a future
-    in-session search can jump straight to a match's window.
+    """Chat-thread turns for a session, oldest first, sourced directly from
+    history_events. offset is a turn ordinal, so a future in-session search can
+    jump straight to a match's window.
+
+    Omitting limit returns the whole session. Paging is opt-in because the
+    callers that cannot page are the ones that read this as a file: the VFS
+    renders sessions/<name>/transcript.md from this route, and a default page
+    size silently truncated every long transcript it served. The viewer asks
+    for a limit explicitly and scrolls through offsets.
 
     No ownership gate: can_read_session is enforced per scope below, so
     another user the session is shared with can read it."""
--- a/backend/services/memory_service.py
+++ b/backend/services/memory_service.py
@@ -344,13 +344,16 @@ async def read_session_events(
 async def read_session_events_page(
     owner_user_id: UUID,
     session_id: str,
-    limit: int,
+    limit: int | None,
     offset: int,
 ) -> tuple[list[dict], int]:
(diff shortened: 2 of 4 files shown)


henry-dowling (changes requested):
This is a fantastic find. 

I'm concerned about the default behavior if no limit is set to be that all transcript events are unceremoniously dumped into the model's context. However, most harnesses have guards against this so probably not that big of a deal in practice. 

Could you please...
(1) merge this
(2) put up a follow-up PR (lower priority) that solves the above problem? An idea- require a param at the API boundary for how many results to request, at the VFS boundary its fine to have a default but then if you get back only one page of a response, loudly tell the model that it's getting back only one page of a response

henry-dowling (approved):

PR #1060, opened 2026-08-18
Add a moderation path for Stash Pages

samzliu:
Stash Pages publishes anonymously, and `DELETE /api/v1/pastes/{slug}` requires the creator's `edit_token` — so **nobody at Stash can remove content someone else published.**

That is how 13 payment-fraud SEO pages have sat on `joinstash.ai` since 2026-08-03: titles like "How to Make a USDT top up virtual card Faster and Easier to Reconcile" and "How to Use no kyc virtual credit cards", one of them linking out to `vccbusiness.com`. They were fully indexable and linked from `/pages`, which is in our sitemap. #1059 makes them `noindex`; this PR is how we actually take them down.

## Why it matters beyond SEO

`noindex` stops search exposure, but the URLs still resolve and still carry `joinstash.ai`'s reputation. Pastebin spam is mostly monetized by dropping links into email and DMs, where a trusted host domain gets past filters — which puts the domain at risk of spam blocklists and Safe Browsing. We send sales and transactional mail from this domain (SPF, DKIM and DMARC are all on the zone), so that is the expensive failure mode, and `noindex` does nothing about it.

## The script

    python -m backend.scripts.delete_pastes <slug> ... --backup out.json

- Prints every matched row and writes a JSON backup **before** deleting, so a mistaken takedown is recoverable.
- Deletes inside a transaction — a run is all-or-nothing.
- Unknown slugs abort the whole run rather than silently deleting the subset that matched.
- Comments cascade with the paste (existing FK).

Parameterized `$1` placeholders, no fallbacks, fails loud on a missing slug — per CLAUDE.md.

## Note

This is a stopgap. If Pages stays up in any form it wants a real admin surface and some rate limiting, otherwise this script gets run every time someone finds the open publish endpoint again.

Touches `backend/**`, so merging cuts a release — which is why it is split out of the `www`-only #1059.

https://claude.ai/code/session_01GxGyzQbUhHjaiRafk1pHf3

### DIFF
--- a/backend/scripts/delete_pastes.py
+++ b/backend/scripts/delete_pastes.py
@@ -1,0 +1,65 @@
+"""Remove pastes by slug.
+
+Stash Pages publishes anonymously and the public delete endpoint requires the
+creator's edit_token, so there is no way for us to take down content someone
+else published — which is how 13 payment-fraud SEO pages sat on the domain in
+August 2026. This is the moderation path until an admin surface exists.
+
+Prints each row and writes a JSON backup before deleting anything, so a
+mistaken takedown can be restored. Comments cascade with the paste.
+
+    python -m backend.scripts.delete_pastes <slug> [<slug> ...] --backup out.json
+"""
+
+import argparse
+import asyncio
+import json
+import sys
+
+from ..database import close_db, get_pool, init_pool
+
+SELECT_SQL = """
+    SELECT id, slug, edit_token, title, content_type, content, visibility,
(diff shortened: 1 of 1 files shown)


samzliu:
Spam takedown is done — the 13 pages are deleted from prod (49 → 36 pastes, live URLs now 404). Closing rather than merging: keeping a standing delete script isn't wanted, and Pages moderation should be solved properly if it comes up again.

PR #1062, opened 2026-08-18
feat: require an explicit event count, and say what the VFS left out

ColemanDuPlessie:
Follow-up to #1058, which was a stopgap. Permanent fix for the same underlying problem.

## Why #1058 wasn't enough

#1058 stopped the silent truncation by removing the server-side default, so omitting `limit` returned the whole session. That traded a silent short read for an **unbounded** one — a 60k-event session now materializes entirely on any `cat` of `sessions/<name>/transcript.md`.

The real defect was never the number 100. It was that **the server was picking a page size that the caller couldn't see**. A short response is indistinguishable from a short session, so no caller could tell whether it had the whole thing — which is exactly how every long transcript came to be rendered cut off, with no marker and exit 0.

## The fix: move the decision to the caller, who is the only one who can disclose it

**API — `limit` is required.**

```python
limit: int = Query(..., ge=1),
```

A bare request is now a 422. `total` and `has_more` already ride along on every response, so a caller that picked a count can report what it missed. `read_session_events_page` goes back to a plain `int`, which deletes the `LIMIT NULL` / `LIMIT ALL` trick #1058 introduced.

**VFS — bounded render, loud disclosure.**

Renders the first `TRANSCRIPT_EVENTS_PAGE = 1000` events (the page size `_load_all_table_rows` already walks tables in). When the session is longer, the body is wrapped **top and bottom**:

```
> ⚠️ TRUNCATED — this file shows the FIRST 1,000 of 4,212 events.
> 3,212 MORE events exist and are NOT below. This transcript is
> INCOMPLETE — do not treat it as the full session.
> Complete session: ./transcript.jsonl
```

Both ends, so a reader who takes the first screenful and one who reads to the end each learn the file is partial. `events.json` carries the same disclosure as `shown` / `total` / `truncated`. `transcript.jsonl` is untouched and stays unpaged — it is what the banner points at, and there is a test asserting it stays whole.

## Deliberate divergence: grep searches more than cat prints

**`grep` searches the WHOLE transcript even though `cat` prints the bounded slice.** A sweep that silently skips the back half of a session is the failure that started this whole thread, and avoiding it is worth more than strict agreement between the two verbs.

The cost is real and is not a bug: grep can match — and print a line number for — text the rendered file does not contain.

```
$ grep -n "deploy" .../transcript.md
.../transcript.md:8431:deploy the thing
$ cat .../transcript.md | wc -l
2104        # line 8431 is not in there
```

So the shell warns whenever it happens:

```
grep: '<path>' was searched in FULL (4,212 events), but the file as rendered
holds only the FIRST 1,000. Matches past event 1,000 are NOT in the file, and
its line numbers will not find them. Complete session: the transcript.jsonl
beside it.
```

Mechanically, nodes gain a `scan_loader` beside the existing `raw_loader` — the same one-node-two-bodies idiom already used to serve a PDF's extracted text to `cat` and its original bytes to `download`. `prefetch(for_scan=True)` warms scan bodies so recursive greps stay concurrent instead of draining transcripts one at a time.

`test_grep_searches_the_whole_transcript_and_says_cat_will_not` pins both halves in one test (grep finds the late match; cat does not contain it) and its docstring says changing this is a product decision, not a cleanup.

## Judgement calls worth a reviewer's eye

- **Scan and render errors share one `error` field.** I considered splitting them so a failed scan can't poison a `cat`, and decided against: the scan's first request is byte-identical to the render's, so a failure that poisons one almost always poisons both, and the model is rebuilt per command so nothing leaks between invocations. One field, per the simplicity rule.
- **`events.json`'s scan body carries the same keys as its rendered body.** Otherwise `grep '"total"' events.json` matched the rendered file but not the searched one.
- **The bottom banner reads "…are NOT below" where nothing is below.** True but mildly odd; I kept one banner string rather than two variants.

## Breaking change

`limit` being required is a breaking API change. Every in-repo caller is accounted for — the frontend always sends it (`frontend/src/lib/api.ts`), and both VFS clients now do. Any out-of-repo script calling this route bare starts getting 422s. That is the intent, but it is worth knowing before merge.

Grep cost also rises: a recursive sweep now drains each transcript in full for the markdown as well as the jsonl. `VfsScanBudget` already meters this and cuts a runaway sweep short with a loud warning.

## Verification

- Full backend suite: **1370 passed**
- CLI suite: **234 passed**
- `ruff check` + `ruff format --check` over `backend/ cli/ stashvfs/`: clean
- Each new test was confirmed to **fail without its mechanism** — reverted the required `limit`, the banner, and the scan path in turn, watched the matching test fail, restored.

New coverage: a short session renders unbannered; a long one states the shortfall at both ends; the grep/cat divergence with its warning; `events.json` disclosure; and `transcript.jsonl` still holding the last event.

> Note for anyone extending these tests: a recursive `grep -r /sessions` also matches the uncapped `transcript.jsonl`, so an assertion meant to exercise `transcript.md` must target that file by path. A `grep -r` assertion passes even when the markdown is completely broken — that fooled me once while writing #1058.


henry-dowling (approved):
Love it, good call on grep

PR #1063, opened 2026-08-19
Redesign the landing page: one system, two audiences

samzliu:
## What

The site now states what we're building and splits into the two audiences that buy it.

| Route | What it is |
| --- | --- |
| `/` | The thesis, then out of the way: four beliefs, two doors, a research index. No product tour. |
| `/internal-agents` | The product page for teams running coding agents. Tabbed quickstart (one prompt / CLI / MCP), import→refine→serve, sharing, ownership, research. |
| `/external-agents` | The production-memory argument, written as documentation with a sidebar and an on-this-page rail. |
| `/blog`, `/contact-sales` | Moved onto the same system. |

## Design

The palette, type, and layout come from a reference pass over the pages this product gets
compared to — Linear, Vercel, Resend, Modal, Cursor, Clay — plus a read of what makes AI-built
pages read as AI-built. The rules that fell out of it: one accent used two or three times a
screen, a short headline over two grey lines, real artifacts instead of diagrams drawn from
rounded rectangles, and hairlines instead of tinted section backgrounds.

Tokens and fonts are repointed in `globals.css` and `layout.tsx`, so every other page inherits
the new brand rather than being stranded on the old one. Chillax and Supreme load from
Fontshare; IBM Plex Mono and Instrument Serif come through `next/font`.

## Removed

- The painted-door `/m/*` variants and their signup flow, which have served their purpose.
- The old homepage sections, and the components only they used: `HeroTerminal`,
  `VisualizationsShowcase`, `EmbeddingProjection3D`, `LiveDemo`.
- `public/hero-funnel.png`, whose only consumer went with them.

## Worth a second look

- **The benchmark claim.** `/internal-agents` says "We have achieved SOTA results on
  LongMemEval and LoCoMo" and `/external-agents` carries a state-of-the-art banner. No numbers
  are printed anywhere. These should be true and citable before this ships.
- The MCP snippet uses `stash-mcp`, the console script from `pyproject.toml`. An earlier draft
  said `stash mcp`, which is not a subcommand.
- `www/DESIGN.md` still describes the old system (Space Grotesk, mono kicker, `#F97316`) and
  needs rewriting to match.

## Checks

`npm run build` passes and generates all routes. `tsc --noEmit` is clean. `npm run lint` has 0
errors and 18 warnings, all pre-existing in files this branch does not touch. Every page was
opened and reviewed in a browser at 1440px.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #1065, opened 2026-08-19
Remove two retired surfaces: the SMB funnel and the public pages pastebin

samzliu:
Follows #1063. Both of these are retired surfaces the redesign left orphaned, so they are
together in one PR.

## 1. The SMB assessment funnel

`/smb` was a chat that produced an "AI Opportunity Snapshot". Nothing links to it and the
campaign is done.

Removes the page, its chat API route (`/api/smb-chat`), its server action, the report template
in `public/smb/`, and the sitemap entry. `Texture` and the `--dot-field` tokens go with it: the
assessment page was their last user once the redesign landed.

## 2. The public pages pastebin

`/pages` was the no-signup pastebin — publish markdown or HTML at `joinstash.ai/pages/{slug}`,
with an agent API over curl and a comment thread.

- **www** — 30 files: the routes under `app/pages/`, the route handlers under `app/api/pages/`,
  the composer, viewer, editor, comments rail, and their helpers.
- **`proxy.ts`** — the rewrite rules that joined those two halves: the POST/PATCH/DELETE agent
  API, the comment paths, and the Accept-header content negotiation that served raw markdown to
  curl and the rendered page to browsers. `withNoIndex` goes too; those routes were its only
  callers.
- **backend** — `routers/pastes.py`, `services/paste_service.py`, the router wiring in
  `main.py`, and `tests/test_pastes.py`.

### What stays, on purpose

The **`pastes` and paste-comment tables** (migration `0112_pastes`). Dropping them destroys
whatever anyone has already published, and keeping them costs nothing. If we want them gone,
that is a separate migration once we are sure nothing published is worth keeping.

**Anyone holding a `joinstash.ai/pages/{slug}` link gets a 404 after this merges.** That is the
intended outcome, but it is the one irreversible part from a visitor's point of view.

## Checks

`npm run build` passes and `tsc --noEmit` is clean. `ruff check` and `ruff format --check` pass
across `backend/` and `cli/`, and `backend/main.py` imports with the router unwired. No
remaining references to `paste_service` or the pastes router anywhere outside the migration.


henry-dowling (approved):
extremely satisfying pr

PR #1067, opened 2026-08-19
Republish two X articles, and fix the SEO the redesign broke

samzliu:
Follows #1063. Two things: the SEO the redesign quietly broke, and two articles brought back
onto our own domain.

## SEO fixes

Three of these are mine, from #1063.

- **`/internal-agents` and `/external-agents` were missing from the sitemap.** The two most
  important new pages were invisible to crawlers.
- **The homepage set its own stale title and description.** Page-level metadata overrides the
  layout's, so `/` still announced "The Company Brain for Your AI Agents and Team" no matter
  what `layout.tsx` said.
- **The `SoftwareApplication` JSON-LD advertised `#pricing` offers** — Free and Pro — pointing
  at a section the redesign removed.
- **The social card was still the old brand**: old paper, old orange, the mono bracket kicker,
  and the old headline. It now matches the page it represents.

`public/logo.png` turned out to be fine — I misread an earlier check from the wrong directory.

## Two articles, republished

Linking out to X does nothing for us: no content on our domain, nothing to index. Both of
these are built out of terms we want to own, so they are now real pages that link into
`/internal-agents` and `/external-agents`:

- `/blog/context-gold-rush` — *The context gold rush: why everyone is building the same thing*
- `/blog/containerizing-memory` — *Containerizing memory: the real barrier to continual
  learning and enterprise AI adoption*

Each keeps a link back to the original X post, and both are in the sitemap and carry
`BlogPosting` markup through the existing `_lib/posts` registry. Handles are rendered as
company names, since a blog post is not a timeline.

## Housekeeping

Every blog post carried its own copy of the pre-redesign header — old nav, `font-black`, the
lot. They all use `SiteHeader` and `SiteFooter` now, which removes the last `font-black` in
the app.

## Checks

`npm run build` passes, `tsc --noEmit` is clean, `npm run lint` reports 0 errors. Both posts
render at 1440px and `/sitemap.xml` emits all four new URLs.

## Still worth doing

No RSS feed, and no per-post OG images — every post currently shares the site card. Say the
word and I'll add both.


henry-dowling (approved):
Bang

PR #1069, opened 2026-08-19
Put the octopus in the hero

samzliu:
The hero's right column has been empty since #1063. This fills it with the app mascot from
`stash_app/app/assets/icon.png`.

- Background flood-filled to transparent from the borders, so the cream card is gone and it
  sits directly on the paper. The flood starts at the edges rather than keying on colour, so
  the highlights inside the drawing survive.
- Trimmed and exported at 900px wide: **69KB webp**.
- The hero becomes a two-column grid at `lg`, stacking on smaller screens.

**Static on purpose.** We tried several animated versions — drawn-on arms, a travelling wave,
a rigged wave using the artwork itself — and they all read as uncanny beside a still headline.
The rigs are kept outside the repo if we want to revisit.

`npm run build` passes, `tsc --noEmit` is clean, lint reports 0 errors.


henry-dowling (approved):
🐙

Issue #1071, opened 2026-08-19
Sessions API: bulk soft-delete by session_id


ischindl:
## Motivation

External integrations that sync with Stash sessions (Fusion's Stash memory backend is the driving case: it soft-deletes the matching Stash session when the corresponding chat session or task is deleted/archived) need to soft-delete **many known `session_id`s exactly, in one request, without a recency window**.

The current surface cannot do this:

- `GET /api/v1/me/sessions` is windowed — `limit` 1..200, `offset` paging, ordered `last_event_at DESC` — and has **no `session_id` filter**. An integration can only scan the recent window and then issue per-row deletes; rows older than the window are silently missed.
- `DELETE /api/v1/me/sessions/{session_row_id}` takes the **row `id` (uuid primary key)**, not the top-level `session_id` string. An integration that only holds `session_id`s must scan the list endpoint to learn the row uuids before it can delete anything.

So today's syncs are bounded (recent window only), multi-request, and carry a documented residual for older sessions.

## Current surface (`backend/routers/sessions.py`, upstream main at b4931afb when filed)

| Route | Handler | Notes |
|---|---|---|
| `GET /api/v1/me/sessions` | `list_my_sessions` (L91) | `limit` 1..200, `offset` paging, no exact `session_id` filter, `last_event_at DESC` |
| `GET /api/v1/me/sessions/{session_id}` | `get_my_session` (L298) | Single-session lookup, scope-bound, read-gated, 404 when absent/unreadable |
| `DELETE /api/v1/me/sessions/{session_row_id}` | `delete_my_session` (L376) | 204 soft-delete; per-row write gate; stamps `deleted_at`/`deleted_by`; audited as `content.session_deleted` |

## Existing-lookup note

Single-session lookup by `session_id` already exists (`GET /me/sessions/{session_id}`, since #667). **This request is specifically the missing *bulk* soft-delete**, not the lookup.

## Proposed contract: `POST /api/v1/me/sessions/soft-delete`

- **Auth/scope:** identical to the sibling `/me/sessions` routes — Bearer auth via `get_current_user`, scope via `get_scope` (`owner_user_id = scope_user_id`; personal scope by default, workspace via `X-Stash-Scope`). No cross-scope leakage, no auth-model change.
- **Body:** `{"session_ids": [str]}` — required, **1..100 items**, each **1..128 chars** (matching `SessionUpsertRequest.session_id`). Server dedupes preserving first-occurrence order.
- **Behavior:**
[...]

PR #1072, opened 2026-08-19
Sessions API: bulk soft-delete by session_id (reference implementation)

ischindl:
## Summary

Reference implementation of #1071: a **bulk soft-delete endpoint for sessions**, addressed by top-level `session_id` strings. Requested by a Fusion integration maintainer — external integrations that sync with Stash sessions need to soft-delete many known `session_id`s exactly, in one request, without a recency window. Today the only path is a windowed list scan (`limit` 1..200, no `session_id` filter) followed by per-row deletes by row uuid: bounded to the recent window and multi-request.

## What this PR adds

`POST /api/v1/me/sessions/soft-delete` in `backend/routers/sessions.py`:

- **Body:** `{"session_ids": [str]}` — 1..100 items, each 1..128 chars (matching `SessionUpsertRequest.session_id`); the server dedupes preserving first-occurrence order.
- **Resolution:** one parameterized `ANY()` query over in-scope rows (`owner_user_id = scope AND session_id = ANY(ids)`), live and trashed rows together — one round-trip.
- **Per-row write gate** via `permission_service.check_access(..., require="write")`; a row the caller lacks write access to is reported `not_found` — existence is never confirmed, the same stance as the canonical lookup.
- **Per-row soft-delete reuses `session_service.delete_session`** — stamps `deleted_at`/`deleted_by` and emits the `content.session_deleted` audit. The service seam is reused, not forked, so every front door keeps leaving the same trail.
- **Always 200 for a valid body:** `{"deleted": [...], "already_deleted": [...], "not_found": [...]}` — each requested id in at most one bucket, request order preserved, idempotent and retry-safe. 401/403 like the sibling routes; 422 for empty/over-cap/wrong-shape bodies.

`backend/tests/test_sessions_bulk_soft_delete.py` — 8 tests covering the contract matrix: happy path (stamps + list removal), idempotent re-delete, `not_found` without blocking siblings in the same request, scope isolation (another user's row stays live), duplicate handling, the 422 boundaries, and audit-trail parity with the per-row delete.

## Non-goals

No schema change, no migration, no new dependencies. No bulk restore/purge. No `session_id` filter on the list endpoint (separate follow-on).

## Test plan

Scoped runs against a scratch Postgres 16 (pgvector) with the full alembic migration chain:

- new suite: 8 passed
- existing `test_session_rename.py` + `test_session_folder_get_or_create.py`: 14 passed

Reference implementation for maintainer review — happy to rework against Stash conventions (route naming, service-layer seam, response shape) if preferred.

### DIFF
--- a/backend/routers/sessions.py
+++ b/backend/routers/sessions.py
@@ -8,11 +8,12 @@ session viewer can ship a Share button without involving the CLI.
 """
 
 import json
+from typing import Annotated
 from urllib.parse import quote
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, File, Form, HTTPException, Query, UploadFile
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, StringConstraints
 
 from ..auth import get_current_user, get_scope
 from ..config import settings
@@ -41,6 +42,13 @@ class SessionUpsertRequest(BaseModel):
     # LEGACY filing lane for installed clients (Heavi's backend foremost):
     # honored when sent, read by nothing new, no default resolved.
     session_folder_id: UUID | None = None
+
+
+class SessionBulkSoftDeleteRequest(BaseModel):
+    session_ids: Annotated[
+        list[Annotated[str, StringConstraints(min_length=1, max_length=128)]],
+        Field(..., min_length=1, max_length=100),
+    ]
 
 
 def _session_app_url(session_id: str) -> str:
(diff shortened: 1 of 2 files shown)


henry-dowling:
Thanks for the PR, we will take a look :)

ischindl:
Rebased onto current `main` (`c56f81c9`). One trivial import conflict in `backend/routers/sessions.py` — main added `from urllib.parse import quote` while this branch added `from typing import Annotated`; unioned, no semantic change to the endpoint. New head: `ac9c8c71`. Verified locally: `backend/tests/test_sessions_bulk_soft_delete.py` — 15 passed (pgvector/pg16). FYI: the only required check here, 'Vercel – stash-web', fails with a `git/authorize` link identically on all our open PRs, so it reads as a Vercel app authorization issue rather than something content-related.

PR #1074, opened 2026-08-19
Tell the user their CLI is stale, instead of guessing why

henry-dowling:
okay so basically we 
(1) just print a stderr to the agent whenever the CLI is out of date on our backend API response.
(2) also try to use pip install if stash was installed via pip
(3) CLI telemetry now has a "CLI version" column so we can better understand how big of a deal CLI going out of date is
(4) look for uv installations in more places than just PATH and `~/.local/bin`, also check Homebrew + Cargo.

# slop
An alternative to #1049, which tried to fix a cause we never established.

On 27 July Sam fixed a bug where an `.html` file inside a directory upload was posted as markdown and rendered as escaped source; it shipped the next day in 0.1.335. On 14 August he hit that same bug himself — so the CLI he was running still predated 0.1.335, while 0.1.364 was current. His machine ran the session-start upgrade throughout. **No record exists of what version he was running, or of where that upgrade stopped working**, because the CLI never compared its version to anything and never reported it.

#1049 responds by replacing the upgrade path with one that covers every way it could have failed at once — Homebrew uv, pyenv shadowing, PEP 668, uv-without-pip, `UV_TOOL_DIR`. Each of those is a hypothesis. This PR does the smallest thing that ends the blindness instead, and leaves the cause to be measured.

## What this does

- Every API response carries `X-Stash-Cli-Latest`, read from the repo's `pyproject.toml` version — the same number `publish.yml` ships to PyPI.
- A CLI that finds itself older prints one line to stderr and carries on:

```
stash: stash 0.1.334 is out of date; the current release is 0.1.365. Run `stash upgrade` to update.
```

- CLI telemetry reports the running version, so "how many installs are stale?" becomes a query.

## What it deliberately does not do

It does not upgrade anything. The plugins' session-start upgrade is untouched and keeps working for everyone it already works for.

An automatic upgrade has to guess at an install kind nobody has measured, and a detached upgrade that fails is invisible in exactly the way this change exists to fix — it announces success it cannot verify, once an hour, forever. A warning is true whatever the cause. Once we can see who is stale, we can find out why on the machines it happens to.

## `stash upgrade`

Fixed, because the warning points at it. It resolved uv off PATH and assumed every install was a uv tool install, so on a pip install under pyenv it upgraded a copy the shell does not run. It now reads the install kind off the running interpreter — a uv tool environment by the receipt uv leaves in it, anything else by its own pip (with PEP 668 flags where required) — refuses an editable checkout, and finds a Homebrew uv.

## Deploy notes

- A pre-release version in `pyproject.toml` raises at import: clients compare plain dotted numbers and would read `0.1.365rc1` as *older*, silently muting the warning fleet-wide.
- `backend/Dockerfile` copies `pyproject.toml`, so a backend that cannot name a release fails at boot rather than going quiet.
- The header is the version the **backend build** was cut from. It equals the newest PyPI release only because a merge to `main` deploys the backend and publishes the CLI from the same commit. If the publish fails and the deploy succeeds, clients are told to chase a release that does not exist.

## Known limit, worth deciding before this ships

The warning goes to stderr. That reaches someone who typed a `stash` command. A notice raised inside the plugin's background transcript upload goes to a process nobody reads. If telemetry shows the fleet is stale in bulk, surfacing this at agent session start is the next step.

## Testing

- `cli/tests/test_release_warning.py` — stale warns, current and newer stay quiet, a missing header says nothing, one line per process across repeated requests, editable checkouts are told to `git pull`; plus the `stash upgrade` install-kind cases (pip by its own interpreter, PEP 668 flags, no pip, uv by receipt with `UV_TOOL_DIR` set, uv install with uv gone).
- `cli/tests/test_release_header_wire.py` — both HTTP clients read the header; the header name is pinned as a literal so renaming the constant cannot pass both suites while breaking every deployed CLI.
- `backend/tests/test_cli_release_header.py` — the header rides every response including 401s, matches the published version, and a pre-release is refused.
- 420 CLI + plugin tests pass; `ruff check` and `ruff format --check` clean on the paths CI lints. The backend tests need a database and were not run locally.
- End-to-end against a mocked backend advertising 0.1.365: a stale install prints the upgrade line once across repeated requests; an editable checkout is told to `git pull`.

---

> [!NOTE]
> **Medium Risk**
> Touches all API responses and shared HTTP clients, but behavior is additive (header + stderr warning) with strict version validation at backend import; upgrade path changes affect user installs directly.
> 
> **Overview**
> Adds **`X-Stash-Cli-Latest`** on every backend response (from `pyproject.toml`, including errors and unauthenticated requests), with import-time failure if the version is not a plain dotted release. The Docker image now copies `pyproject.toml` so deploys cannot boot without a comparable release number.
> 
> The CLI and plugin HTTP clients call **`stashai.release.note_latest`** on each response so a behind install prints **one** stderr line (`stash upgrade` or `git pull` for editable checkouts). **`stash upgrade`** now picks **uv tool** vs **pip** (PEP 668 flags when needed) from the running interpreter instead of always assuming uv. CLI telemetry events include **`version`** for fleet staleness queries.
> 
> The Claude plugin **`ensure_cli.sh`** widens **uv** discovery (Homebrew/cargo paths) and aligns the reinstall message with the shared installer string. Version bump to **0.1.366**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 990b0a650bbc5dfe17ac6a4c9621867b9a41e001. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -38,6 +38,8 @@ COPY docs/skills/ ./docs/skills/
 COPY stashvfs/ ./stashvfs/
 COPY backend/entrypoint.sh ./entrypoint.sh
 COPY alembic.ini ./alembic.ini
+# The CLI release number clients are told about (backend/cli_release.py).
+COPY pyproject.toml ./pyproject.toml
 
 RUN mkdir -p /app/.secrets && \
     chmod 700 /app/.secrets && \
--- a/backend/cli_release.py
+++ b/backend/cli_release.py
@@ -1,0 +1,38 @@
+"""The stash CLI release clients are told about.
+
+Sourced from the repo's pyproject version — the same number publish.yml pushes
+to PyPI — so cutting a release stays a one-line edit. Served on every response
+as X-Stash-Cli-Latest; a CLI that finds itself older says so (see
+stashai/release.py). Note this is the version of the build this backend was cut
+from, which is the newest published release only because a merge to main
+deploys the backend and publishes the CLI from the same commit.
+
+A missing or malformed version raises at import: a backend that cannot name a
+release clients can compare would silently stop warning anyone, which is the
+failure this path exists to end.
+"""
+
+import re
+import tomllib
+from pathlib import Path
+
+LATEST_VERSION_HEADER = "X-Stash-Cli-Latest"
+
+_PYPROJECT = Path(__file__).resolve().parent.parent / "pyproject.toml"
+
+
+def _read_release(pyproject: Path) -> str:
+    with pyproject.open("rb") as f:
+        release = tomllib.load(f)["project"]["version"]
+    # A pre-release ("0.1.365rc1") is PEP-440-valid and would publish fine, but
+    # clients compare plain dotted numbers and would read it as *older* —
+    # silently muting the warning until the next plain release.
+    if not re.fullmatch(r"\d+(\.\d+)*", release):
+        raise ValueError(
+            f"pyproject version {release!r} is not a plain dotted release; "
+            "clients cannot compare it and would stop noticing new releases."
+        )
+    return release
+
+
+LATEST_CLI_VERSION: str = _read_release(_PYPROJECT)
--- a/backend/main.py
+++ b/backend/main.py
@@ -8,6 +8,7 @@ from slowapi.errors import RateLimitExceeded
 from slowapi.middleware import SlowAPIMiddleware
 from starlette.responses import JSONResponse
 
+from . import cli_release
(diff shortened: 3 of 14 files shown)


ColemanDuPlessie (approved):
Looks good, although I skimmed the changes to tests, which comprise a majority of the lines changed in this PR (let me know if you'd rather I read them more thoroughly).

henry-dowling:
I would not! You should definitely skim test code.

PR #1097, opened 2026-08-28
Release 0.1.368: self-host compose fix + GHCR pins + release contract test (no pi in images)

ischindl:
# Release 0.1.368 — self-host images (frontend + backend, no pi), compose orphan fix, pins promoted

**Audience:** self-hosters. joinstash.ai is unaffected by the compose changes (hosted prod runs on Render from the Dockerfiles, not compose) — but remember **any merge to `main` also auto-deploys Render prod**, PyPI, and GHCR.

## What this PR ships
- **Semantic-search / MCP capability** (already committed by you on this branch: `ffbb9115` import + `ac5a79f6` STAS-113 fixture fix): `/me/events/semantic-search`, `/me/pages/semantic-search`, `/{table_id}/rows/semantic-search`, the `stash_search` / `stash_session_search` / `get_stash_info` MCP tools, and the local sentence-transformers embedding provider for self-hosts (`openai → huggingface → local` auto chain). Capability proven pre-push: 8 + 39 + 15 tests green (semantic-search, backend MCP, CLI MCP server).
- **Compose orphan fix:** `docker-compose.local.yml` carried a `collab: ports: 3458` override for the service #982 deleted. Compose *creates* override-only services and then rejects them — the documented pair died with `service "collab" has neither an image nor a build context specified`. The block is removed; `config -q` now exits 0. The `backend`/`frontend`/`caddy-profiles` overrides survive untouched.
- **One version commit** (`chore: bump stash release to 0.1.368 …`): `pyproject.toml` 0.1.367 → **0.1.368** and all five GHCR pins (backend/worker/heavy-worker/beat ×4 + frontend, previously `0.1.365`) → `0.1.368`, plus a CHANGELOG bullet.
- **A release contract test** (`backend/tests/test_docker_release_contract.py`): every local override service must exist in prod (generalizes the collab bug class), neither compose file may mention collab/3458, **GHCR pins must equal the CLI version** (bare tags), the two plugin manifests must agree with each other, and no pi reference may appear in either Dockerfile.

## What this PR deliberately does NOT do
- **No pi in the images.** Operator directive (2026-08-28): *“The Docker image must not install pi at all. pi is not for self-hosting.”* Both Dockerfiles are **byte-identical to `origin/main`** in this diff. No server-side pi harness/provider/local-exec — that all stays on PR #1088's founder-gated lineage.
[...]

PR #1088, opened 2026-08-23
STAS-117: land the STAS-098 pi-harness delta (local-model cloud agent) on main

ischindl:
Adds a "Local model" option for the cloud chat agent. Instead of a Claude, Codex, or OpenRouter key, users can connect their own OpenAI-compatible endpoint (Ollama, vLLM, etc.) and the agent runs that model through the pi CLI on their cloud computer.

Setup is Settings → Cloud agent model → Local model: enter a base URL and model name (API key optional). The endpoint has to be reachable from the user's cloud computer — for a laptop that usually means a tunnel (cloudflared, ngrok, ssh). If it's unreachable, the turn fails with a clear error instead of falling back to anything.

This ports the STAS-098 change, which is already landed and fully tested on our internal main. The branch is that same 27-file diff, cherry-picked off a fresh main, verified green locally (1454 backend tests, 307 frontend, ruff clean). It needs to be on main before the Docker release images (STAS-101), since the release base is this repo's main.

Two housekeeping items: GitHub is asking a maintainer to approve-and-run the first CI run for this fork PR (actions/runs/32781047863), and the Vercel check is waiting on a one-time fork authorization (not a required check for main).

Internal ref: STAS-117 (cherry-pick of STAS-098).

One pre-existing red flagged by our internal merge spec is N/A for this PR: 4 tests in `backend/tests/test_memory_semantic_search.py` (`sessions.last_event_at` NotNullViolationError) were recorded red on the origin-lineage. That test file exists only in our internal main lineage — it is absent from origin/main (verified at base 1cbda10c) and not part of this PR's 27-file diff — so those tests do not run in this PR's CI and there is nothing to port; the internal one-line fixture fix (STAS-113) has no origin-lineage target.

PR #1099, opened 2026-08-29
fix(STAS-144): remove dead collab-service leftovers from docs and self-host config

ischindl:
## What

Delete the last live-looking references to the collab sidecar that PR #982 removed, in the four non-`docs/` surfaces, and pin the invariant with a static guard test so this rot class cannot silently return.

The collab service (`collab/`, `backend/routers/collab.py`, the Yjs/Hocuspocus server; table dropped by migration `0182`) no longer exists, but four surfaces still described it as running — and one of them made the **documented self-host command fail outright**:

```
service "collab" has neither an image nor a build context specified: invalid compose project
```

`README.md` tells a self-hoster to run exactly the failing pair of compose files.

## Changes

| Surface | Change |
|---|---|
| `ARCHITECTURE.md` | drop the `backend/routers/collab.py` bullet (file absent) |
| `render.yaml` | drop `collab` from the hosted-prod list in the header comment; the only service this file *defines* is `aspose-pptx` |
| `.env.example` | delete the "Live Collaboration" block (`ws://localhost:3458`, the nonexistent `/collab` Caddy route) and `BACKEND_URL` |
| `docker-compose.local.yml` | delete the orphaned `collab:` ports override |
| `backend/tests/test_no_dead_service_references.py` | **new** static guard: 3 tests |

### `BACKEND_URL` is deleted, not renamed

It is the only key in `.env.example` with **zero consumers** repo-wide (case-insensitive search over `backend`, `cli`, `frontend/src`, `www`, `plugins`, `docker-compose.prod.yml`). Its own comment said "used by the collab sidecar". `docker-compose.prod.yml` passes the differently-named `BACKEND_INTERNAL_URL`, which is a different variable with many consumers. Per repo rules there is no alias, shim, or fallback.

## Deliberate overlap with STAS-101 / PR #1097

- `87f87d3e` (STAS-101, on `fusion/stas-101`) deletes the **same 4 lines** of `docker-compose.local.yml`. Our result is **byte-identical** to `git show 87f87d3e:docker-compose.local.yml`, so the two branches converge at merge (identical deletions auto-resolve). We did not rebase onto or cherry-pick it — this branch stays on `origin/main`.
- Rationale for not waiting: #1097 was **OPEN** (`mergedAt: null`, head `d09d2eb8`) and `87f87d3e` is not an ancestor of `origin/main`, while the self-host command is broken *now*.
- #1097 also adds `test_compose_files_do_not_mention_collab`, which asserts the same compose invariant. The overlap is intentional; **once #1097 merges, its contract test subsumes the compose half of ours** and the compose entries can be dropped from our file.
- Our other two guards (architecture router existence, dead `BACKEND_URL`) cover surfaces #1097 does not touch.

## Deliberate non-coupling with STAS-143

The guard test **does not scan `docs/`**: `docs/security-readiness.md:87,98` is owned by STAS-143, and asserting on a surface this PR does not clean would go RED whichever of the two independent PRs merges first. Likewise the router guard is one-directional (docs → disk); the reverse ("every router is documented") would be red today for unrelated reasons — 28 real routers are undocumented.

## Explicitly untouched (not dead-service references)

`CHANGELOG.md:44,47` (historical release notes), `backend/tests/test_page_events.py` docstrings documenting the removal, migrations `0064`/`0182` (immutable history), `frontend/package-lock.json` `prosemirror-collab` (a real npm library), `collaboration`/`collaborations` prose, `docs/security-readiness.md`, `.github/` (zero collab mentions).

## Verification

**Symptom red → green** (the `config -q` run follows README's sequence, i.e. after `cp .env.example .env`; compose validates merged services before env-file, which is why the collab error masked the env check before):

```
$ docker compose -f docker-compose.prod.yml -f docker-compose.local.yml config -q
- service "collab" has neither an image nor a build context specified   exit=1   (before)
                                                                        exit=0   (after)
$ docker compose ... config --services
postgres redis backend worker heavy-worker beat frontend      # caddy is declared but
                                                              # profile-gated by design
```

**Guard test red → green** (the red transcript names exactly the offenders this PR deletes):

```
before: 3 failed
  AssertionError: ARCHITECTURE.md names routers that do not exist: ['collab.py']
  AssertionError: deleted collab service is still advertised at:
    ['ARCHITECTURE.md:81 ...', 'render.yaml:6 ...', '.env.example:54 ...',
     '.env.example:55 ...', '.env.example:58 ...', 'docker-compose.local.yml:18 ...',
     'docker-compose.local.yml:20 - "3458:3458"']
  AssertionError: .env.example declares unread variables: ['BACKEND_URL=http://localhost:3456']
after:  3 passed
```

**Grep invariant:**

```
$ grep -rn --exclude-dir=node_modules --exclude-dir=.git -iw collab \
    ARCHITECTURE.md render.yaml .env.example docker-compose.local.yml
(no matches)          # CHANGELOG collab count unchanged at 2, same as origin/main
```

**Lint:** `ruff check backend/ cli/ && ruff format --check backend/ cli/` with the CI-pinned `ruff==0.15.6` → "All checks passed!" + "625 files already formatted". No Python module and no TypeScript changed, so ruff + the scoped pytest run are the complete gates.

## Related

- **STAS-143** — same rot class, owns `docs/security-readiness.md`. Merges in either order with this PR, by the design above.
- **STAS-145** — the `www/package.json` dead editor dependency block (`@hocuspocus/provider`, `yjs`, the `@tiptap/*` suite with zero imports in `www/`) is already filed there; not duplicated.

🤖 Generated with [Fusion](https://runfusion.ai)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,14 +50,6 @@ CORS_ORIGINS=http://localhost:3457,http://localhost:3456
 # Must be an HTTPS origin without a path when managed Auth0 is enabled.
 # APP_BASE_URL=https://app.yourcompany.com
 
-# ── Live Collaboration ───────────────────────
-# The frontend defaults to ws://localhost:3458 in local dev and /collab on
-# deployed same-origin installs. Keep Caddy's /collab route when using the
-# prebuilt self-host frontend image.
-
-# Used by the collab sidecar to authorize users through FastAPI.
-BACKEND_URL=http://localhost:3456
-
 # ── Embeddings ──────────────────────────────────────────────────
 # Provider: openai | huggingface | local | none | auto (default).
 # "auto" detects from available keys: OPENAI_API_KEY → openai, HF_TOKEN → huggingface, else → local.
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -78,7 +78,6 @@ of the current architecture. Privacy is mediated by Skills.
 - `backend/routers/tables.py`: structured table CRUD + row search.
 - `backend/routers/shares.py`: person-to-person folder shares.
 - `backend/routers/sources.py`: GitHub / Google Drive / Notion OAuth + imports.
-- `backend/routers/collab.py`: Yjs WebSocket sidecar for live page editing.
 - `backend/routers/trash.py`: soft-delete listing + restore/purge.
 
 Object-level privacy is enforced inline in each router that returns a
--- a/backend/tests/test_no_dead_service_references.py
+++ b/backend/tests/test_no_dead_service_references.py
@@ -1,0 +1,88 @@
+"""The collab sidecar is gone (PR #982); the repo must stop describing it.
+
+The Yjs/Hocuspocus collaboration service was deleted — `collab/` and
+`backend/routers/collab.py` no longer exist and migration 0182 dropped its
+table — yet the architecture document and the self-hosting files kept naming
+it. One of those stale references (`collab:` in docker-compose.local.yml) made
+the exact self-host command README.md tells you to run fail outright with
+`service "collab" has neither an image nor a build context specified`.
+
+These are static source-level guards in the shape of
+test_cli_release_header.py: they read repo files as plain text and assert the
+invariant, so no one can quietly re-add a service description for something
+that was deleted. docs/ is deliberately NOT scanned — docs/security-readiness.md
+is owned by STAS-143, and asserting on a surface this task does not clean would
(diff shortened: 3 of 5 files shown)


ischindl:
Closed per operator directive 2026-08-30: push freeze during #1088 dogfooding; additionally #1101's branch was cut off the superseded fold/b-2026-08-27 lineage (35 unmerged commits under 4 real ones). Branches kept. Cleanup line will be re-specified from main after #1088 lands.

PR #1104, opened 2026-09-01
Create end users by hand from the developer console

henry-dowling:
This makes it possible to add end users by hand in the developer console as an admin convenience.

# Slop
## Why

End users only came into existence when a product backend uploaded a session with a new `user_id`, so a developer could not set a user up ahead of onboarding — assign them a Drive folder, seed their wiki — without faking an upload. Found while dogfooding the developer console.

## What changed

- `POST /api/v1/me/users` creates an end user by hand on the workspace scope. Takes `user_id` (the developer's own id, same field uploads assert) plus an optional `name`; reuses `get_or_create_end_user`, so the wiki folder is created exactly as the upload path does it.
- A duplicate `user_id` is a 409 — a manual duplicate is more likely a typo than an intended no-op — and the first upload later carrying the id lands on the same user rather than forking a second one.
- The console Users page gains a "…" menu above the table; "Add a user" opens the form (id + optional name, inline error kept for correction), which closes on X or a successful add.
- Drive-by cleanup in its own commit: deleted a stray PNG committed at the repo root under the literal filename `--full-page` (an agent-browser screenshot arg mishap).

## How it was verified

- `pytest backend/tests/test_developer_platform.py` — 23 passed, including three new tests: manual create then upload joins the same user, duplicate 409s, non-workspace scope 400s
- `npm test` — 61 files, 309 tests passed (6 new component cases: menu gating, create-and-close, blank name omitted, server error keeps form open, disabled until typed, X closes)
- `ruff check` / `ruff format --check` / `npm run lint` — clean (10 pre-existing warnings in unrelated files)
- Clicked through live: created a user from the form, saw the 409 inline on a repeat, opened the created user's detail page

https://claude.ai/code/session_01Rq2QegXVmXAm4ptrTN8RFU

---

> [!NOTE]
> **Medium Risk**
> Delete runs multi-step transactional cleanup on folders, pages, and files; mistakes are destructive for private wikis though scope checks and tests cover isolation and history retention.
> 
> **Overview**
> Developers can **provision end users before the first upload** and **manage them from the Users console**, not only when sessions arrive with a new `user_id`.
> 
> **API:** `POST /api/v1/me/users` takes the developer’s `user_id` plus optional `name`, creates the user and private wiki via the same path as uploads, and returns **409** if that external id already exists so later uploads attach to the same row. **`DELETE /api/v1/me/users/{id}`** removes the identity and private wiki (wiki files soft-deleted, pages/tables in that tree removed); sessions and other uploaded files stay in the workspace with **no user assigned**, and the external id can be recreated. **`PATCH`** now applies only fields present in the body (no silent clears via `null`) and rejects empty patches.
> 
> **Console:** A **“…” menu** adds **Add a user** (`AddUserControls`) and per-row **Rename / Delete** (`UserActions`) with a confirmation that explains retention. **`createUser` / `deleteUser`** client helpers wire the new endpoints; the user table layout separates navigation from wiki toggle and actions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c372292c5ac0a2635dd8df1f403c19bb95c597f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samzliu (approved):
Can we delete users as well/ do we have full CRUD in the developer console too?

henry-dowling:
@samzliu good point! actioned

PR #1105, opened 2026-09-01
add embeddings viz job to the heavy queue

henry-dowling:
We have two separate queues of tasks in celery. There's one for easy tasks, and one for hard tasks that could block the queue for a while. This PR moves embeddings viz computation to the "hard tasks" queue.

## Why

The 2 GiB default Celery worker OOM-restarted twice because each visualization precompute UMAP fit permanently raised a prefork child’s RSS by roughly 280–530 MB. Production runs four default-worker children, so their combined memory can cross the container limit without any child reaching the existing 800 MB recycle threshold.

## What changed

- Route `backend.tasks.viz.precompute` to the 8 GiB, two-child heavy worker
- Preserve the invariant that cadence-sensitive Beat dispatchers stay on the default queue
- Pin visualization precompute as the sole intentionally heavy Beat task in the routing regression test

## What to look at

The behavioral change is only the Celery route. Visualization refreshes may wait behind other heavy work, but their cache TTL is six hours and Beat checks every five minutes, so they are not latency-sensitive.

## How it was verified

- `pytest -q -o addopts='' --noconftest backend/tests/test_celery_routing.py` — 3 passed
- `ruff check backend/ cli/`
- Confirmed the configured route resolves to `{"queue": "heavy"}`

The normal focused pytest command could not initialize the repository-wide fixture because no local Postgres is running in this worktree; the routing tests were rerun without `conftest.py` because they do not access the database.

---

> [!NOTE]
> **Low Risk**
> Single Celery route change with regression tests; main tradeoff is viz precompute latency behind other heavy tasks, not correctness or security.
> 
> **Overview**
> Moves **`backend.tasks.viz.precompute`** from the default Celery queue to **`heavy`**, so memory-heavy UMAP work runs on the larger worker pool instead of contributing to OOM on the 2 GiB default workers.
> 
> Beat still fires viz precompute every five minutes; only the execution queue changes. Precompute may queue behind other heavy jobs, which is acceptable given the six-hour cache TTL.
> 
> The routing test is updated to treat viz precompute as the **sole** beat task allowed on `heavy`, replacing the previous rule that no beat task could be heavy. A small comment tweak in `celery_app.py` clarifies the default-queue policy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03a0220284f4b3fe4fbd2f68bf4018d0736caca8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/celery_app.py
+++ b/backend/celery_app.py
@@ -52,8 +52,8 @@ celery.conf.update(
     # Two queues, split by task weight. Anything that can hold a worker slot
     # for minutes — subprocess extractions, Playwright renders, exports,
     # source crawls, bulk URL fetches, headless agent runs — routes to
-    # "heavy" and gets its own worker pool. Everything else (every beat
-    # sweep, every cheap bounded task) stays on "default", which therefore
+    # "heavy" and gets its own worker pool. Everything else (cheap beat
+    # sweeps and bounded tasks) stays on "default", which therefore
     # never stalls: cadence-sensitive tasks like X token keep-fresh (its
     # 45-min refresh_margin assumes a tick roughly every 30 min) get their
     # guarantee without being enumerated. Interactive agent replies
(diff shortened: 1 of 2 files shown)


samzliu (approved):
I did not realize that we had two celery queues

henry-dowling:
@samzliu I added it sometime earlier this month since our curator was getting blocked by lower-priority scheduled tasks like embeddings

PR #1106, opened 2026-09-01
Hand a user files and sessions from the developer console

henry-dowling:
This PR makes it possible to manually add a file or a session to the developer console, as a developer convenience

# Slop

> **Stacked on #1102** — based on `codex/org-drive-source-ui` so the diff shows only this work; retarget to `main` lands automatically when 1102 merges. Together with #1102 (Drive sources) and #1104 (manual user creation), this completes manual setup of an end user from the console.

## Why

The user detail page could only display a user's material — files and sessions had to arrive through the product backend's API calls. A developer setting a user up ahead of onboarding (dogfooding feedback) had no way to hand them a file or a session transcript.

## What changed

- **Files**: an "Upload a file" button in the Files section. Rides the pre-existing `user_id` form field on `POST /me/files` (which turned out to have no test — added one); the file row is stamped `end_user_id`, so only that user's agent sees it.
- **Sessions**: a transcript upload hidden behind a "…" menu — sessions normally stream from the product backend, so uploading one by hand is the unexpected path. The form takes a `.jsonl` file (session id prefills from the filename stem, editable) and an agent name.
- **Backend**: the transcript endpoint gains an optional `user_id`. Two deliberate constraints: the end user must already exist (unknown id → 400), and a session born into a different user — or none — refuses the upload with "already belongs to…" rather than migrating. `end_user_id` is set-at-insert-only (it is the session's privacy boundary), so the user is resolved before any session row is written.
- A "skipped" upload (session already has events) surfaces inline as "Nothing imported: …" instead of pretending it imported.

## How it was verified

- `pytest backend/tests/test_developer_platform.py backend/tests/test_transcripts.py` — 44 passed, including new tests: transcript files the session under the user, unknown user 400s, cannot move a session between users, file upload stamps `end_user_id`
- `npm test` — 63 files, 313 tests passed (7 new component cases across both controls)
- `ruff check` / `ruff format --check` / `npm run lint` — clean (10 pre-existing warnings in unrelated files)
- Clicked through live: uploaded a file and a 2-event transcript for a user; both appeared in their sections and both DB rows are stamped to that user; the transcript's own timestamps date the session

https://claude.ai/code/session_01Rq2QegXVmXAm4ptrTN8RFU

---

> [!NOTE]
> **Medium Risk**
> Changes touch end-user privacy boundaries (session `end_user_id` at insert and cross-user rejection); behavior is guarded by new tests but transcript upload is a new code path for `user_id`.
> 
> **Overview**
> Developers can **manually attach files and session transcripts to a specific end user** from the user detail page in the developer console, without going through the product backend.
> 
> **Backend:** `POST /me/transcripts` gains an optional `user_id` form field. When set, the handler resolves the end user (must already exist → 400 if unknown), runs the same cross-user session guard as event batch uploads, and passes `end_user_id` into `upsert_session` so the session is born in that user's privacy boundary. Re-uploading a transcript for a session that already belongs to another user is rejected instead of migrating.
> 
> **Frontend:** New **Upload a file** control on the Files section (uses existing `user_id` on `POST /me/files`). Sessions get a **… → Upload a session transcript** flow with `.jsonl`, session id (prefilled from filename), and agent name; skipped imports show inline (*Nothing imported: …*) rather than refreshing the list.
> 
> **Client API:** `uploadFile` and `uploadTranscript` accept an optional `userId` that is sent as `user_id` in the multipart form.
> 
> **Tests:** Integration coverage for user-scoped transcript upload, unknown user, session user collision, and file `end_user_id` stamping; component tests for both upload controls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec46893b97093d8b78101582b41b07df811b1962. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/backend/routers/transcripts.py
+++ b/backend/routers/transcripts.py
@@ -21,6 +21,7 @@ from fastapi.responses import PlainTextResponse
 from ..auth import get_current_user, get_scope
 from ..database import get_pool
 from ..services import (
+    end_user_service,
     memory_service,
     security_audit_service,
     session_service,
@@ -52,6 +53,9 @@ async def upload_transcript(
     session_id: str = Form(...),
     agent_name: str = Form(...),
     cwd: str | None = Form(None),
+    # External Multiplayer: file the session under this end user (the
+    # developer's own id for them). The user must already exist.
+    user_id: str | None = Form(None),
     # LEGACY filing lane for installed clients: honored when sent.
     session_folder_id: UUID | None = Form(None),
     replace: bool = Form(False),
@@ -74,6 +78,20 @@ async def upload_transcript(
     body = await file.read()
     if len(body) > MAX_TRANSCRIPT_SIZE:
         raise HTTPException(status_code=413, detail="Transcript too large (max 50 MB)")
+
+    # end_user_id is set at session insert only (the user a session is born
+    # into is its privacy boundary), so the end user must resolve before any
+    # session row is written — and a session already born into a different
+    # user (or none) must refuse the upload rather than migrate.
+    end_user = None
+    if user_id is not None:
+        try:
+            end_user = await end_user_service.resolve_end_user_for_scope(owner_user_id, user_id)
+            await memory_service.reject_cross_user_sessions(
+                owner_user_id, [{"session_id": session_id, "user_id": user_id}]
+            )
+        except ValueError as e:
+            raise HTTPException(status_code=400, detail=str(e))
 
     pool = get_pool()
     existing = await pool.fetchval(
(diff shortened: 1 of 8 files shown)


samzliu (approved):
What's the difference between a user's files and a user's wiki? Should the latter just be a folder?

henry-dowling:
files = uploaded PDFs that the model can use to make the memory wiki
wiki = memory wiki, derived from raw sources like sessions and files

does that seem clear to you? i agree it's a bit confusing

samzliu:
@henry-dowling - I was asking more about the backend implementation since it seems like files are treated differently b/c they're a tag rather than a folder?

Not sure if it's confusing to the user in the UI though

